PR #72, opened 2014-12-09
Use node url module to parse db url

### LINKED ISSUE #71: use_env_variable rejects valid postgres urls
The url parsing around lib/tasks/db.js:33 rejects urls has a couple of problems:
1. it rejects some valid urls, for example urls without username/password
2. it fails with a strange message when a URL is rejected:
```
[10:03:30] Starting 'db:migrate'...
[10:03:30] 'db:migrate' errored after 483 μs Cannot read property '6' of null
```

It seems like it might be better to use the [node built-in url module] (http://nodejs.org/docs/latest/api/url.html#url_url_parse_urlstr_parsequerystring_slashesdenotehost) instead of a regular expression for parsing.

davidmfoley:
Resolves #71 

IMPORTANT NOTE:

I tested this with some postgres URLs locally but there might be cases where either:

1. Some task (that I don't use) in sequelize is relying on some old behavior 
2. Someone else might use a database url that is not handled by this change

So: someone who knows the internals of sequelize and the cli a lot better than I do should review this.

I didn't see any existing tests for this so I didn't write any new ones.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var path      = require("path");
+var url       = require("url");
 var helpers   = require(path.resolve(__dirname, "..", "helpers"));
 var args      = require("yargs").argv;
 var _         = require("lodash");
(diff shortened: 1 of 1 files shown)


theanimal666:
Please, please merge - just implemented the same fix for my own project and wanted to file a pull request (https://github.com/theanimal666/cli/commit/e0ff831eb5b6a81083c9e0a81ed86ed52fb10735) an then saw that there is this pull request with the same code already pending 

kind regards,
Chris

sdepold:
There you go :)

sdepold:
Please try master and let me know if it works correctly. Will release a new version of the CLI later

theanimal666:
thank you very very much - will let you know when i tested it.

theanimal666:
works perfect - thanks!

sdepold:
great :)

sdepold:
Just released 1.0.3

PR #86, opened 2015-01-23
allow using specific harmony flags

alubbe:
--harmony enables all features, so I prefer to enable just what I need - generators


sdepold:
:+1:

sdepold:
Released as 1.0.6

PR #88, opened 2015-01-27
Fixes issue where `tryToMigrateFromOldScheme` function was crashing mysteriously.

nathschmidt:
If the file names in the migration did not contain a timestamp (either no timestamp or timestamp not terminated by a dash). Simply added a check on the result of filename.match, if it's null it's assumed not to be a migration. Otherwise all is apples.

Tested on my own projects, however I haven't run it against the unit tests as there doesn't seem to be tests for the eloquently named `tryToMigrateFromOldScheme` function.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -251,8 +251,12 @@ function tryToMigrateFromOldSchema () {
       var files      = fs.readdirSync(helpers.path.getMigrationsPath());
 
       return files.filter(function (file) {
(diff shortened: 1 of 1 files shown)


sdepold:
Tests are green. Green is good. So this PR must be good either. :) Thanks a lot.

sdepold:
Just released 1.0.7

nathschmidt:
Sweet.  Thanks for updating.  Was a fun afternoon tracking that down :)

PR #99, opened 2015-02-13
Fix db:migrate:old_schema with sequelize >= 2.0.0

adamhooper:
In Sequelize 2.0.0, `sequelize.query()` defaults to `type: 'RAW'` instead of `type: 'SELECT'`.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -245,7 +245,7 @@ function tryToMigrateFromOldSchema () {
     return queryInterface.renameTable("SequelizeMeta", "SequelizeMetaBackup")
     .then(function () {
       var sql = queryInterface.QueryGenerator.selectQuery("SequelizeMetaBackup");
-      return sequelize.query(sql, null, { raw: true });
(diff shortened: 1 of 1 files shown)


adamhooper:
I consider this fix critical: without it, anybody who runs `sequelize db:migrate:old_schema` will end up with a completely empty `SequelizeMeta` table.

janmeier:
Agreed, this is critical. However we also have to ensure that the error does not apply the other way around (using spread with an older sequelize version. 

Can we signal with a codependency or something like that that the next version of the cli only works with sequelize 2.0.0 and later

sdepold:
So this change is not backwards compatible?

mickhansen:
@sdepold it should definitely be BC

sdepold:
OK great. That's what I assumed from reading the change. I added a follow-up PR: #100

sdepold:
The new one fixes the quotes and brings some additional lovely builds to travis including multiple versions of sequelize.

sdepold:
I'm closing this one. Please continue commenting on #100

sdepold:
Just released in 1.3.0

PR #100, opened 2015-02-13
Follow-up of #99

sdepold:
This PR

- is a follow-up of #99 
- fixes new default for `sequelize.query` in 2.0.0
- adds builds for sequelize @ 1.7.10, 2.0.0-rc8 and 2.0.0 final

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ before_script:
   - "mysql -e 'create database sequelize_test;'"
   - "psql -c 'create database sequelize_test;' -U postgres"
   - npm install js2coffee@$(echo $JS2COFFEE)
+  - npm install sequelize@$(echo $SEQUELIZE)
 
 script:
   - "node_modules/.bin/gulp"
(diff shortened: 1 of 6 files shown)


sdepold:
Alright so it seems that the CLI is not compatible with sequelize 1.7.x and 2.0 stable :(

mickhansen:
A change for 2.0 was raw query handling, so if you have raw queries multiple places we need to go through that.

sdepold:
Alright Umzug is not 1.7 compatible. I'm on it in this very moment

sdepold:
mickhansen is there any kind of property in sequelize which will tell me if it is 1.x or 2.x ?

sdepold:
To be more concrete: I try to find a way to find out if I need to use `destroy({ where: {foo:'bar'} })` or `destroy({ foo: 'bar' })`

mickhansen:
Hmm no i don't think we include the version anywhere, not sure though.
Might have to parse 'sequelize/package.json'

sdepold:
Is `sequelize.daoFactoryManager` 1.x only?

sdepold:
OK apparently `sequelize.modelManager` is 2.x only.

mickhansen:
Yeah `modelManager` was a 2.0 rename.

sdepold:
OK umzug seems to work

sdepold:
1.7 compatibility is in place now

sdepold:
alright. such green. very nice

sdepold:
just released 1.3.0

Issue #101, opened 2015-02-19
Feature: Allow umzug JSON storage


itoed:
Hi folks,

I was looking at the [task that interfaces with umzug](https://github.com/sequelize/cli/blob/87b7b836b39ac7bda001f17f5e25f538e895c266/lib/tasks/db.js#L181) and I noticed the storage is hardcoded to Sequelize.

I have a requirement where the migrations tracking data cannot be on the database.  I'm now trying to use Umzug without the sequelize-cli.

So, it would be really nice if you could allow the option to configure umzug JSON storage.

Thank you!

sdepold:
Hmmm..... Well.... That should be pretty easy. Would you like to provide a PR for that?

itoed:
A PR has been created (#103).  Please let me know if there is anything else I can do so that sequelize cli can have this feature.

sdepold:
this is done right?

itoed:
Yes, everything seems fine with v1.4.0 now.  Thank you so much for all your help!

PR #103, opened 2015-02-23
Add option for JSON storage

itoed:
Please refer to Feature Request #101.

~~This commit introduces options `--storage` and `--storage-path` to tell Umzug to use JSON storage instead of MySQL table SequelizeMeta.~~

This commit introduces configuration options options `migrationStorage` and `migrationStoragePath` to tell Umzug to use JSON storage instead of MySQL table SequelizeMeta.

### DIFF
--- a/lib/helpers/umzug-helper.js
+++ b/lib/helpers/umzug-helper.js
@@ -1,0 +1,28 @@
+"use strict";
+
+var path    = require("path");
+var _       = require("lodash");
+var helpers = require(__dirname);
+
+module.exports = {
+  getStorage: function () {
+    return helpers.config.readConfig().migrationStorage || "sequelize";
+  },
+
+  getStoragePath: function () {
+    return helpers.config.readConfig().migrationStoragePath ||
+      path.join(process.cwd(), "sequelize-meta.json");
+  },
+
+  getStorageOptions: function (extraOptions) {
+    var options = {};
+
+    if (this.getStorage() === "json") {
+      options.path = this.getStoragePath();
+    }
+
+    _.assign(options, extraOptions);
+
+    return options;
+  }
+};
--- a/test/db/migrate-json.test.js
+++ b/test/db/migrate-json.test.js
@@ -1,0 +1,123 @@
+"use strict";
+
+var expect    = require("expect.js");
+var Support   = require(__dirname + "/../support");
+var helpers   = require(__dirname + "/../support/helpers");
+var gulp      = require("gulp");
+var fs        = require("fs");
+var _         = require("lodash");
+
+([
+  "db:migrate",
+  "db:migrate --migrations-path migrations",
+  "--migrations-path migrations db:migrate",
+  "db:migrate --migrations-path ./migrations",
+  "db:migrate --migrations-path ./migrations/",
+  "db:migrate --coffee",
+  "db:migrate --config ../../support/tmp/config/config.json",
+  "db:migrate --config " + Support.resolveSupportPath("tmp", "config", "config.json"),
+  "db:migrate --config ../../support/tmp/config/config.js"
+]).forEach(function(flag) {
+  var prepare = function(callback, options) {
+    options = _.assign({ config: {} }, options || {});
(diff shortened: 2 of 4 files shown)


sdepold:
Thanks a lot :) Will check the PR in the evening!

test/db/migrate-json.test.js
```diff
@@ -0,0 +1,53 @@
...
+      .pipe(helpers.runCli(flag, { pipeStdout: true }))
+      .pipe(helpers.teardown(callback));
+  };
+
+  describe(Support.getTestDialectTeaser(flag), function() {
+    it("creates a sequelize-meta.json file", function(done) {
+      var jsonFile = Support.resolveSupportPath("tmp", "sequelize-meta.json");
+
+      prepare(function() {
+        expect(fs.statSync(jsonFile).isFile()).to.be(true);
```
sdepold: Would it make sense to check the content of the file?

lib/helpers/umzug-helper.js
```diff
@@ -0,0 +1,29 @@
+"use strict";
+
+var args = require("yargs").argv;
+var path = require("path");
+var _    = require("lodash");
+
+module.exports = {
+  getStorageMethod: function () {
+    return args.storage || "sequelize";
```
sdepold: What is your opinion on supporting a config file?

sdepold: This function will return the content of the config file https://github.com/sequelize/cli/blob/master/lib/helpers/config-helper.js#L63

itoed: Yes, adding the two flags to the existing config file makes sense.  I'll work on that.

About your comment on checking the contents of the `sequelize-meta.json` file:  I could not understand how to use `readFile` defined in `test/support/helpers`.

Can you give me a hint on how I can do an `expect` using `readFile`?  I'm not familiar with `through2`.  Or should I just use `fs`?

Thank you for your feedback so far!

sdepold: I'm fine with using `fs`. tbh I kinda regret introducing gulp to the tests. Anyways, this is how you can use the readFile pattern:

```
 gulp
          .src("dir")
          .pipe(helpers.readFile("sequelize-meta.json"))
          .pipe(helpers.expect(function(stdout) {
              expect(stdout).to.contain("up: function(migration, DataTypes, done) {");
          })
          .pipe(helpers.teardown(done));
```

itoed: Thanks for the tip!  I saw this after I added a test with `fs`.  I will add the config file reading to this PR soon.

itoed:
I've now changed the umzug helper to read the migration storage options from the sequelize configuration file only. E.g.

```
{
  "production": {
    "username": "root",
    "password": null,
    "database": "database_production",
    "host": "127.0.0.1",
    "dialect": "mysql",
    "migrationStorage": "json",
    "migrationStoragePath": "sequelize-meta.json"
  }
}
```

In hindsight, I think declaring a migration storage strategy via command line arguments is a bad idea.

In my npm scripts, I had `node_modules/.bin/sequelize db:migrate --storage='json'`.  But after moving them into the sequelize configuration file I feel that it's the only place where that should be allowed.  If somebody switches strategies accidentally, you could end up with unintended schema changes.

Please let me know if you agree, and how I can continue this pull request.

Tell me when you are ready to merge and I'll squash all my commits.

Thanks for your attention!

sdepold:
I think I agree :+1: I will make a final review in the evening or tomorrow. :) Thanks for your work!

sdepold:
Cool :) Can you please extend the `README.md`?

sdepold:
The rest looks good :)

itoed:
Okay, I have documented the usage in the `README.md`.  Please keep in mind that in the near future it might be good to cut some of the tests to speed up the build.  Thank you!

sdepold:
Thanks a lot.

sdepold:
Released as 1.4.0

PR #108, opened 2015-03-13
Fixed second try to resolve path to gulp

streuspeicher:
The fallback to resolve gulp below current path did not work on windows because unix shell script was executed instead of gulp.js


sdepold:
Released as 1.3.2

PR #111, opened 2015-03-23
addded db:migrate:undo:all to undo all migrations at once

Americas:
Simple task to undo all migrations, instead of going one by one.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -60,6 +60,29 @@ module.exports = {
     }
   },
 
+  "db:migrate:undo:all": {
+    descriptions: {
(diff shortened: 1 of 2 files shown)


sdepold:
Awesome! Exactly what I was just about to implement :)

sdepold:
Do you think you could try to add some tests for that feature?

Americas:
How do I get your tests to use postgres instead of mysql? I have created a test based on the undo test, it should grab 2 migrations, test if they are applied, run undo:all, and test if they are both gone.

sdepold:
nice :) thank you!

sdepold:
merged and released as 1.5.0

PR #126, opened 2015-04-21
support ES6+ syntax by babel

fundon:


fundon:
@sdepold I submited a new and clean PR.

lib/assets/migrations/create-table.es6.js
```diff
@@ -0,0 +1,37 @@
...
+          allowNull: false,
+          type: Sequelize.DATE
+        },
+
+        updatedAt: {
+          allowNull: false,
+          type: Sequelize.DATE
+        }
+      })
+      .done(done);
```
sdepold: This line has to be removed

lib/assets/models/index.js
```diff
@@ -5,7 +5,7 @@ var path      = require('path');
 var Sequelize = require('sequelize');
 var basename  = path.basename(module.filename);
 var env       = process.env.NODE_ENV || 'development';
-var config    = require(<%= configFile %>)[env];
```
sdepold: This change will break https://github.com/sequelize/cli/blob/22f208e7fda83268a8ae003b1074a12573befd83/lib/helpers/init-helper.js#L72

sdepold: And the fact that the build is still green makes me think, that we should add a test for that

sdepold:
Is it a common thing to add `es6` to the file extension? I think it's odd :)

test/support/assets/migrations/20111117063700-createPerson.es6.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+export default {
+
```
sdepold: Is this line break personal taste or a common thing in ES6 land? I'd prefer the removal of the padding lines in a block.

sdepold:
Good job :) I will take another final look later :)

PR #139, opened 2015-06-08
Document hidden --url property

brianhartsock:
I was perusing the code and noticed the --url command line option. It is very useful for my use cases and I thought documenting it might be valued. I have only used it for a few commands thus far, but it is proving useful.

Thoughts?


sdepold:
Awesome :) didn't even know about it's existence :o

PR #149, opened 2015-07-08
Hotfix/use env variable

### LINKED ISSUE #135: Document use_env_variable and database.json
Environment variables appear to be supported, but I was unable to find anything in the documentation and just happened to be searching issues and saw another example. This should be far easier to find in my opinion. On this note the entire database.json should be documented

```json
  "production": {
    "use_env_variable": "PRODUCTION_ENV_DATABASE_URL"
  }
}
```

The CLI is a great feature and the help sections are very good, but again I came up empty here.
```bash
$ sequelize help:init:config

Sequelize [Node: 0.10.30, CLI: 1.7.0, ORM: 2.1.3]

Loaded configuration file "server/config/database.json".
COMMANDS
    sequelize init:config -- Initializes the configuration.

DESCRIPTION
    Initializes the configuration.

OPTIONS
    --env              The environment to run the command in. Default: development
    --coffee           Enables coffee script support. Default: false
    --config           The path to the config file. Default: config/config.json
    --options-path     The path to a JSON file with additional options. Default: none
    --migrations-path  The path to the migrations folder. Default: migrations
    --models-path      The path to the models folder.Default: models
```

sdepold:
Fixes #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,32 +23,38 @@ $ sequelize [--HARMONY-FLAGS]
 
 
 ```
-Sequelize [CLI: v0.0.4, ORM: v1.7.5]
+Sequelize [Node: 0.10.36, CLI: 1.7.0, ORM: 2.0.2]
 
 
 Usage
   sequelize [task]
 
 Available tasks
-  db:migrate        Run pending migrations.
-  db:migrate:undo   Revert the last migration run.
-  help              Display this help text. Aliases: h
-  init              Initializes the project.
-  init:config       Initializes the configuration.
-  init:migrations   Initializes the migrations.
-  init:models       Initializes the models.
-  migration:create  Generates a new migration file. Aliases: migration:generate
-  version           Prints the version number. Aliases: v
+  db:migrate             Run pending migrations.
+  db:migrate:old_schema  Update legacy migration table
+  db:migrate:undo        Revert the last migration run.
+  db:migrate:undo:all    Revert all migrations ran.
+  help                   Display this help text. Aliases: h
+  init                   Initializes the project.
+  init:config            Initializes the configuration.
+  init:migrations        Initializes the migrations.
+  init:models            Initializes the models.
+  migration:create       Generates a new migration file. Aliases: migration:generate
+  model:create           Generates a model and its migration. Aliases: model:generate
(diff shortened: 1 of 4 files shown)


vikeen:
:+1: 

This issue mostly stemmed from the fact that the config is the best choice (IMO) to configure sequelize, but the actually docs didn't specify everything it could do. This is a big step forward in understanding it's power, but is anything else missing too?

README.md
```diff
@@ -23,35 +23,40 @@ $ sequelize [--HARMONY-FLAGS]
...
-  init              Initializes the project.
-  init:config       Initializes the configuration.
-  init:migrations   Initializes the migrations.
-  init:models       Initializes the models.
-  migration:create  Generates a new migration file. Aliases: migration:generate
-  version           Prints the version number. Aliases: v
+  db:migrate             Run pending migrations.
+  db:migrate:old_schema  Update legacy migration table
+  db:migrate:undo        Revert the last migration run.
+  db:migrate:undo:all    Revert all migrations ran.
```
JHouk: Typo here: 'ran' = 'run'.  

From a practicality standpoint, I think most migration histories occasionally hit migrations that can't be rolled back. Not sure how realistic it is to be able to truly support undo:all.

vikeen: I've never had a problem migrating back through any number of iterations. That is after all the point of using a migration tool. Have you had issues?

JHouk: When migrations result in dropping of legacy tables or fields, that tends to be crossing the rubicon. :)

JHouk:
Thanks @sdepold!

MrOrz:
:+1: lgtm

README.md
sdepold: Well it heavily depends on the use cases. While I was working for companies that invested quite some time in proper migrations which where perfectly undoable at any given point in time, I also saw the strategy of using one-way migrations only.

Anyways, this feature is useful in development for sure. 

Thanks for the heads up about the typo. I sort of think it makes the way it is but maybe I'm wrong. Feel free to provide a PR which changes it here: https://github.com/sequelize/cli/blob/master/lib/tasks/db.js#L65

PR #150, opened 2015-07-10
Fix args to sequelize.query

jsha:


sdepold:
Closed in favor of #153

Issue #152, opened 2015-07-14
db:migrate:old_schema broken since sequelize version 3.0.0


peterjwest:
This command relies on `sequelize.query` (`/lib/tasks/db.js:273`) which was changed in sequelize 3.0.0

sdepold:
thanks for the heads up

rjmunro:
So I guess the workaround is to upgrade to umzug before you upgrade sequelize?

sdepold:
I will fix the actual problem today ig guess

peterjwest:
@rjmunro yes, that would work

sdepold:
#153 will fix this issue

sdepold:
Just released version v1.7.2 of the CLI. Please check if it is working properly now :)

peterjwest:
Works perfectly,  you're :star2:THE BEST:star2:

sdepold:
great :+1:

PR #153, opened 2015-07-15
Hotfix/sequelize 3

sdepold:


test/support/assets/migrations/20111117063700-createPerson.coffee
```diff
@@ -7,8 +7,8 @@ module.exports =
           type: DataTypes.BOOLEAN
           defaultValue: false
           allowNull: false
-    .complete done
+    .done () -> done()
```
mickhansen: could also use `nodeify` to keep the old way of doing it, the new way won't capture errors

sdepold:
Fixes #152 and replaces #150

test/support/assets/migrations/20111117063700-createPerson.coffee
sdepold: good point. need to change that

Issue #110, opened 2015-03-19
db:migrate returns status code 0 when migration fails


renou:
Migration failures on db level are not handled properly and the command returns success code 0 and prints

Using environment 'development'.
== 20150318142814-init: migrating =======
Possibly unhandled SequelizeDatabaseError: type "enum_Group_status" already exists

sdepold:
It would be nice if you could create a test as a pull request that shows the failing behavior. Otherwise I'll try to take a look at it at some point.

renou:
```javascript
var sh = require('execSync');
var result = sh.exec('node_modules/.bin/sequelize db:migrate');
if(result.code || result.stdout.indexOf('SequelizeDatabaseError') > -1) {
  logger.error('has SequelizeDatabaseError but return code is 0');
  throw new Error(result.stdout);
} else {
  applauseLogger.info(result.stdout);
}
```

theazureshadow:
+1. We need to fail deployments that didn't migrate correctly, and without an exit code our deployment scripts can't tell that they failed.

arturhtml:
+1

sorin:
+1

sdepold:
fixed in 1.7.4

PR #161, opened 2015-07-27
Return exit code 1 when migrations failed

johanneswuerbach:
Fixes https://github.com/sequelize/cli/issues/110


test/db/migrate.test.js
```diff
@@ -19,6 +19,8 @@ var _         = require('lodash');
 ]).forEach(function (flag) {
   var prepare = function (callback, options) {
     options = _.assign({ config: {} }, options || {});
+    options.cli = options.cli || {};
```
sdepold: Why not `options.cli = _.extend({ pipeStdout: true }, options.cli)`?

sdepold:
:clap:

PR #163, opened 2015-08-01
Environment variable support in config file

szantner:
For the config variables 'host', 'port', 'username', 'password', 'database' adds the possibility of using environment variables with the following syntax:

"username": {"ENV": "DB_USERNAME"}

Helps in integration with services like AWS Elastic Beanstalk

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -171,6 +171,12 @@ function getSequelizeInstance () {
     console.log(e.message);
     process.exit(1);
   }
+	
+	['host', 'port', 'username', 'password', 'database'].map(function (key) {
+		if(config[key] && config[key]['ENV']) {
+			config[key] = process.env[config[key]['ENV']];
+		}
+	});
 
   _.forEach(config, function (value, key) {
     if (['database', 'username', 'password'].indexOf(key) === -1) {


kerihenare:
In this case you could just use the URL parsing: `{ 'url': 'mysql://' + process.env.MYSQL_USER + ':' + process.env.MYSQL_PASSWORD + '@' + process.env.MYSQL_HOST + '/' + process.env.MYSQL_DATABASE }`

szantner:
Could you please clarify a bit more? As far as I understand, if you use the CLI to migrate you can either use a config file, the command line option --url or a composed connection string as an environment variable. Probably I am missing something here.

kerihenare:
I use a `.sequelizerc` file that looks something like:
```
module.exports = {
    'migrations-path': './migrations',
    'models-path': './models',
    'url': 'mysql://' + process.env.MYSQL_USER + ':' + process.env.MYSQL_PASSWORD + '@' + process.env.MYSQL_HOST + '/' + process.env.MYSQL_DATABASE
};
```

szantner:
No I see, thanks. It seems to be a nice workaround, in the meantime I think it would be more elegant to have a direct support to do that in the config file. First of all if you want to remain flexible, you have to implement the whole connection string composition code taking into account all the possible connection options. Also I don't know what happens to the rest of the settings in the config file if you use a connection string in .sequelizerc (I would assume the framework ignores it in that case, but I am not sure) 
Of course if there will be no support for individual environment variables in the config file, this seems to be the solution.

kerihenare:
@szantner Perhaps a better technique might be to allow for the config file to be either a JSON file or a JS file. So if it finds a JS file instead, it'll look something like:

```
module.exports = {
    development: {
        username: 'root',
        password: null,
        database: 'database_development',
        host: '127.0.0.1',
        dialect: 'mysql'
    },
    test: {
        username: 'root',
        password: null,
        database: 'database_test',
        host: '127.0.0.1',
        dialect: 'mysql'
    },
    production: {
        username: 'root',
        password: null,
        database: 'database_production',
        host: '127.0.0.1',
        dialect: 'mysql'
    }
};
```

Also allowing:
```
module.exports = {
    username: 'root',
    password: null,
    database: 'database_development',
    host: '127.0.0.1',
    dialect: 'mysql'
};
```

Then you could also just have:
```
module.exports = {
    username: process.env.MYSQL_USERNAME,
    password: process.env.MYSQL_PASSWORD,
    database: process.env.MYSQL_DATABASE,
    host: process.env.MYSQL_HOST,
    dialect: 'mysql'
};
```

szantner:
@kerihenare I think both solutions can work for the env variable problem, but probably your suggestion gives more flexibility in the future. I guess if you could submit a pull request with your solution, that might speed up the process.

sdepold:
OK so 2 things about this comment: https://github.com/sequelize/cli/pull/163#issuecomment-128219930

1)
Would it help if we check for a `config/config.js` in addition to `config/config.json`?

2)
This is already possible when using the sequelizerc file (though it's hacky imho).
So if you have `.sequelizerc` file which looks like this

```
'use strict';

var path = require('path');

module.exports = {
  'config': path.resolve('config', 'config.js')
};
```

then the CLI will load the `config/config.js` file in which you can do what was suggested.

dnlsandiego:
I think it would really help checking for `config/config.js` in addition to `config/config.json`. I think the CLI would be more flexible that way. May I ask why is it hacky to configure `.sequelizerc` to look for `config.js` instead of `.json`? Since after encountering this discussion, I went with the `config.js` route so I can use the CLI with environment specific settings placed in `env` and am wondering if there is any major drawbacks to it.

sdepold:
Well it's not really hacky but it makes sense to me that the CLI would check for `config.js` by default.

szantner:
Doesn't seem too hacky for me either and works perfectly. But if you decide to support reading config.js directly - that is even better.

sdepold:
Hey guys. Could you give this a try? https://github.com/sequelize/cli/pull/194

I'm a bit worried that people are storing a configuration in config.js already. So I guess an upcoming release will be a major bump.

sdepold:
Please check v2.0.0

manuelfink:
Has one of you deployed to aws? It seems to me that .env is not available conatainer_commands?

Shefali25:
I have multiple database configurations. So my development environment is defined in this manner:
`module.exports = {
    development: {
        db1: {},
        db2: {}....
    },
    test: {},
    production: {}
};`
where it connects to db1 by default. How can I manage this?

PR #164, opened 2015-08-03
Mask password when outputting parsed URL

kerihenare:
Don't output password to CLI (for security).


sdepold:
Good change. Guess we should add a test for that. I can take care of it if you want to

sdepold:
was just released as 1.8.1

PR #175, opened 2015-08-28
[Changed] Cascade package resolution

sdepold:
Try to load sequelize first via the native require.resolve and then
with the node-resolve module.


sdepold:
@rfink this should fix your problem

PR #114, opened 2015-03-31
RFC: db:seed and db:unseed

Americas:
Based on Issue #109, I have done what you suggested and added commands for seeding and unseeding. It is not production ready, and there are no tests, but I wanted you to check it out and tell me your thoughts. I realize some parts are a little bit hammered on, especially the umzug-helper.js code.

PR #176, opened 2015-08-28
db:seed and db:unseed

ataube:
This is a working PR based on #114. Merged with master and fixed the tests. 

Many thanks to @Americas for the initial work. Lets get it shipped...


ataube:
Damn, there is an issue with mysql. I guess it's the countTable() function, using proper parameters should fix that...

sdepold:
:cool: are you going to fix it?

test/support/assets/seeders/20111117063700-seedPerson.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+var nodeify = require('nodeify');
+
+module.exports = {
+  up: function (migration, DataTypes, done) {
+    nodeify(migration.bulkInsert('Person', [{
```
janmeier: No need to `nodify`, you should be able to just return the promise

ataube:
Cool thanks for the feedback, I will fix it. Any suggestions for the countTable() function. Is it the right approach to fix the mysql error if I use proper query replacements? 
http://docs.sequelizejs.com/en/latest/docs/raw-queries/

janmeier:
pass `{ type: sequelize.QueryTypes.SELECT}`, then you will get the result directly, instead of an array

ataube:
ok guys, this PR is ready to be merged. There is still one test failing, same as master ;-)

@janmeier I still use nodify just about consistency, all other seeders/migrations use the same pattern. If this is still an issue for you, I would suggest opening a new issue for that.

sdepold:
I restarted one of the builds

lib/helpers/umzug-helper.js
```diff
@@ -4,32 +4,43 @@ var path    = require('path');
...
+  getStorage: function (type) {
+    return this.getStorageOption(type + 'Storage', 'sequelize');
   },
 
-  getStoragePath: function () {
-    var fallbackPath = path.join(process.cwd(), 'sequelize-meta.json');
+  getStoragePath: function (type) {
+    var fallbackPath = path.join(process.cwd(), storageJsonName[type]);
+
+    console.log(fallbackPath);
```
sdepold: This is a leftover I guess?

test/support/assets/seeders/20111117063700-seedPerson.js
sdepold: IIRC it contains `nodify` to support the old versions of sequelize (the CLI basically supports all version of sequelize).

sdepold:
@Americas @ataube :ok_hand: awesome stuff!

lib/assets/seeders/skeleton.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    /*
+      Add altering commands here.
+      Return a promise to correctly handle asynchronicity.
+
+      Example:
+      return queryInterface.createTable('users', { id: Sequelize.INTEGER });
```
sdepold: Should this do something with bulkInsert?

lib/assets/seeders/seed-table.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
```
sdepold: Where are we using this file?

sdepold:
functionality wise it seems to work nicely :+1:

lib/helpers/umzug-helper.js
ataube: good spot, this one shouldn't be there...

lib/assets/seeders/skeleton.js
ataube: correct, the example should be seeder specific...

lib/assets/seeders/seed-table.js
ataube: This seems like a copy past mistake, I couldn't find any reference to this file => deleting it

sdepold:
:cool: I'm going to squash the commits, fix a tiny typo and merge it manually

sdepold:
Just released as 1.9.0. Thanks a lot for your work!

durazzocl:
I've just test it and it seems you should update your template.

QueryInterface.prototype.bulkInsert = function(tableName, records, options, attributes) {
  options.type = QueryTypes.INSERT;
  var sql = this.QueryGenerator.bulkInsertQuery(tableName, records, options, attributes);
  return this.sequelize.query(sql, options);
};

This function implies that "options" is not null. But your template indicates that options is "optionnal"

Example:
      return queryInterface.bulkInsert('Person', [{
        name: 'John Doe',
        isBetaMember: false
      }]);

So without it, it throw an exception.

Except that, it works perfectly. Thank you very much. It's a very usefull feature

sdepold:
Meh, good point. It's a pity that options is not optional -.-

PR #186, opened 2015-09-22
--underscore to create migrations with 'created_at' and 'updated_at'

Americas:
Needed a break from thinking, this feature was simple to do, so I just did it.

### DIFF
--- a/lib/tasks/migration.js
+++ b/lib/tasks/migration.js
@@ -13,7 +13,7 @@ module.exports = {
       options: {
         '--name': 'Defines the name of the migration. ' +
           clc.blueBright('Default: unnamed-migration'),
-        '--underscored': 'Set timestamps to snake_case'
+        '--underscored': 'Use snake case for the timestamp\'s attribute names'
       }
     },
 
--- a/lib/tasks/migration.js
+++ b/lib/tasks/migration.js
@@ -12,7 +12,8 @@ module.exports = {
       'short': 'Generates a new migration file.',
       options: {
         '--name': 'Defines the name of the migration. ' +
-          clc.blueBright('Default: unnamed-migration')
+          clc.blueBright('Default: unnamed-migration'),
+        '--underscored': 'Set timestamps to snake_case'
       }
     },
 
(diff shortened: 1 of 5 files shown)


sdepold:
Thanks :) Do you think you want to add tests?

Americas:
added it for model:create, I can add it for migration:create, but it will follow the same paths. Will add it tomorrow.

test/model/create.test.js
```diff
@@ -128,47 +128,63 @@ var _         = require('lodash');
...
-                  '        type: Sequelize.DATE',
-                  '      }'
-                ].join('\n')))
-                .pipe(helpers.ensureContent('});'))
-                .pipe(helpers.ensureContent('.dropTable(\'Users\')'))
-                .pipe(helpers.teardown(done));
+          ([
+            { underscored: true, createdAt: 'created_at', updatedAt: 'updated_at'},
+            { underscored: false, createdAt: 'createdAt', updatedAt: 'updatedAt'}
+          ]).forEach(function (attrUnd) {
```
sdepold: What is `attrUnd`? :D

Americas: its the ATTRibute for UNDerscored, you already have an attributes declared above, so I had to use something else

lib/tasks/migration.js
```diff
@@ -12,7 +12,8 @@ module.exports = {
       'short': 'Generates a new migration file.',
       options: {
         '--name': 'Defines the name of the migration. ' +
-          clc.blueBright('Default: unnamed-migration')
+          clc.blueBright('Default: unnamed-migration'),
+        '--underscored': 'Set timestamps to snake_case'
```
sdepold: maybe "Use snake case for the timestamp's attribute names" ?

sdepold:
test looks good

lib/tasks/migration.js
Americas: Will change it when I get to my PC :wink:

puzanov:
Who can merge? We need this very much

sdepold:
@puzanov did you try out the branch? is it working properly?

sdepold:
works nicely

sdepold:
Released as 2.1.0

PR #194, opened 2015-09-27
[Added] Read configuration from config.js if available

sdepold:
This PR makes the CLI read the configuration from `config/config.js` if available. This way it is possible to have dynamic content in the configuration file (e.g. you could read environment variables).

### DIFF
--- a/bin/sequelize
+++ b/bin/sequelize
@@ -3,7 +3,6 @@
 'use strict';
 
 var spawn   = require('child_process').spawn;
-var fs      = require('fs');
 var path    = require('path');
 var args    = process.argv.slice(2);
 var yargs   = require('yargs');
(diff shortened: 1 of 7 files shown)


sdepold:
Fixes https://github.com/sequelize/cli/pull/163

bassarisse:
Just wanted to point out that `fs.existsSync` is currently deprecated ;)

sdepold:
> Just wanted to point out that fs.existsSync is currently deprecated ;)

nooooo! :D Thanks for the hint. I recall reading something about that.

sdepold:
`fs.accessSync(path, fs.R_OK)` is the way to go these days?

bassarisse:
Yes, and we now have to check for exceptions, no more boolean value returned...

bassarisse:
Sorry for the delay, I just tested this, it worked beautifully =)

sdepold:
:cool: Will update the PR to use accessSync for newer node versions and release during the week.

sdepold:
I really have to fix those annoying tests at some point :(

@bassarisse can you review my usage of fs.accessSync ?

sdepold:
Released as v2.0.0

bassarisse:
Good to know! Sorry for not responding, was on vacation.

PR #196, opened 2015-09-28
Change logging config to console.log function when set

cogpie:
I noticed a deprecation warning from the Sequelize library, resulting from setting the logging config in the Sequelise-cli option to true: 
https://github.com/sequelize/sequelize/commit/00399231d04ef617bdfb29826c4de3097f5bf553

As your code uses console.log elsewhere, this just compliments that (assuming similar semantics for the logging option as it is passed in)

Although, could so something more elaborate with converting a string to the function eg.
```javascript
        if (_.isString(this.config.logging)) {
          var funcPath = this.config.logging.split(".");
          var func = global;
          while (funcPath[0] && func) { func = func[funcPath[0]]; funcPath.shift(); }
          this.config.logging = (_.isFunction(func))? func : false;
        }
```
using the config option:
```json
...
   logging: "console.log"
...
```

### DIFF
--- a/lib/helpers/config-helper.js
+++ b/lib/helpers/config-helper.js
@@ -97,7 +97,7 @@ module.exports = {
 
         // The Sequelize library needs a function passed in to its logging option
         if (this.config.logging) {
-          this.config.logging = console.log
+          this.config.logging = console.log;
         }
         
         this.config = this.config[env];
--- a/lib/helpers/config-helper.js
+++ b/lib/helpers/config-helper.js
@@ -99,7 +99,7 @@ module.exports = {
         if (this.config.logging) {
           this.config.logging = console.log;
         }
-        
+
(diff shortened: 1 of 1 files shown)


sdepold:
I totally ran into that issue as well one day and noticed this problem too. I guess this change makes sense. OTOH I'm not entirely sure where we are actually using the `config.logging`. Guess it's just for the migration part in which we are instantiating sequelize.

sdepold:
Can you please add a semicolon?

cogpie:
No problem, all done. Although I'm not sure I'll have time to repair the many test errors I saw pop up. The config.logging parameter is passed to the Sequelize library, and this is handy to see what it has been up to for dubugging. I guess, in theory, the cli could use it too - but for now this will make you compatible with any changes to Sequelize atm.

Issue #154, opened 2015-07-15
How do I run the migration with SSL enabled?


STRd6:
I'm trying to run a migration on a remote postgres host that requires SSL, passing `--url=...` on the command line. How do I enable SSL?

dmcquay:
I have the same question, but I am using the config file. I added "ssl": true, but it still does not connect via SSL.

ben-ng:
+1

ryanwalters:
+1

sdepold:
Hm I have a bit of hard times to answer this question, as I don't have an SSL enabled endpoint available.

ryanwalters:
You can spin one up for free on Heroku.

sdepold:
:cool: Will try to do that ASAP (which is probably one evening during this week)

ryanwalters:
The SSL Heroku connection string looks something like:

    postgres://username:password@ec2-12-34-56-78.compute-1.amazonaws.com:5432/database?ssl=true

The `?ssl=true` comes from [Heroku's Docs](https://devcenter.heroku.com/articles/heroku-postgresql#connecting-in-node-js). It is required when connecting remotely.

So back to `sequelize`, I try to pass this value into `--url=...`, including `?ssl=true` and I get an error like:

    Unable to connect to database: SequelizeConnectionError: no pg_hba.conf entry for host "12.34.56.78", user "username", database "database?ssl=true", SSL off

You'll notice that the database value that is being parsed **includes the query parameter `?ssl=true`** and **SSL is not being enabled**.

Digging in a little bit, I see that [`urlStringToConfigHash`](https://github.com/sequelize/cli/blob/master/lib/helpers/config-helper.js#L139-L160) ignores the query string.

It seems like two things might help here:

1. Remove query string from database
1. If `ssl=true` is being passed, enable it

        // Tried this just for testing, and it doesn't work. How do we enable SSL?
        if (urlParts.query) {
            result = _.assign(result, {
                ssl: true
            });
        }

Any thoughts?

icemilo:
My PR has been merged, could you test if specifying dialectOptions will work for postgres?

jfishman:
:warning: The dialectOptions PR #220 allows users to specify the requisite ssl options, fulfilling this issue as far as the ssl feature request component, but the `--url` parameter parsing is still broken.  A flag named '--url' is reasonably expected to support valid URL forms, with or without query parameters. The database component is specified by the URI path, not including any part of the query.

[...]

PR #220, opened 2015-11-07
Add options for dialectOptions

icemilo:
There's a test for dialectOptions but when options are loaded from config file, dialectOptions get ignored. In order to create ssl connection to ssl enabled endpoints, we need to pass dialectOptions to mysql driver.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -276,6 +276,12 @@ function getSequelizeInstance () {
         });
       }
     }
+
+    if (key === 'dialectOptions') {
+      options = _.assign(options, {
+        dialectOptions: value
+      });
+    }
   });
 
   options = _.assign({ logging: logMigrator }, options);


sdepold:
You are probably the first person on this planet who managed to create a green build :smile_cat:  Thanks for your work.

sdepold:
@icemilo is this correct? https://github.com/sequelize/cli#dialect-specific-options

icemilo:
@sdepold yes that should do it :)

sdepold:
Thanks for your feedback :)

sdepold:
@icemilo Do you think this PR fixes https://github.com/sequelize/cli/issues/154 too ?

icemilo:
@sdepold I have no experience for postgres but this PR should pass the options to dialects, so I'm guessing it should work for postgres as well

PR #223, opened 2015-11-14
Attribute Parse (WIP)

topleft:
I have created a set of parsing functions that would allow for field attributes to be specified via the command line on `create:model`. Work still needs to be done on the templating of both the model and migrations for this to work. And of course updating the tests. I would like to continue to hack at it, but I am having trouble modifying the tests. I may have a templating solution for the model on another branch, but won't know until the tests run. Any help/advice would be awesome!


topleft:
[repo with tests](https://github.com/topleft/cli-parse-tests) for these parse functions

topleft:
@sdepold can I get some feedback on this before I re-write ALL the tests, please. Thanks!

sdepold:
@topleft so this allows:

```
sequelize model:create --name foo --attributes name:[type:string,required:true] age:[type:integer, required: false]
```
?

Can we keep the backwards compatibility and support 
```
sequelize model:create --name foo --attributes name:[type:string,required:true] age:integer
```

?

topleft:
I think so. Let me rework the parse function.

topleft:
@sdepold 
I have adjusted the parsing functions to accept the the current command format. I think this is a good start to allowing for more specific attribute definition, thus saving a lot of work for the user in updating two files (model and migration). There are several issues/questions that need to be addressed.

### Backwards Compatibility
* The tests are not passing because the object I am returning is a different format and the templates (model and migration) can't use them 
* Using the current template would require two sets of parsing functions

* The output format in the model file: 
```
{ fieldName: DataTypes.TYPE }
```
  * Desired Output
```
{ fieldName: {
     type: DataTypes.TYPE
  }
}
```
  * My parsing functions return the object in this desired format 

### Current Hangups
##### Lodash Template
* model.js
* migration.js

##### Tests
* as this parsing is fundamental, many of the tests are touched by this change

_I am interested in taking this further with some help. What do you think?_

sdepold:
I'm OK with changing the format of the output if the previous CLI interface is backwards compatible (`sequelize model:create --name foo --attributes name:[type:string,required:true] age:integer`).

sushantdhiman:

PR #228, opened 2015-11-20
Filtering everything onf the go

rutaihwa:
Gain a tiny advantage of by doing all filtering at the same time, since the code is run synchronously.


sdepold:
:cool:

PR #243, opened 2015-12-26
db.Model instead of db.model

oss92:
When I define my model as such

```
  var Model = sequelize.define('model', attributes, {
    ...
```

It is more conventional and consistent to reference my model as `db.Model` and not `db.model` and still the naming conventions for foreign keys and so do not screw up.

### DIFF
--- a/lib/assets/models/index.js
+++ b/lib/assets/models/index.js
@@ -21,7 +21,8 @@ fs
   })
   .forEach(function(file) {
     var model = sequelize['import'](path.join(__dirname, file));
-    db[model.name] = model;
+    var modelName = model.name.charAt(0).toUpperCase() + model.name.slice(1)
+    db[modelName] = model;
   });
 
 Object.keys(db).forEach(function(modelName) {


sdepold:
It's a pity that no tests are breaking. dafuq.

lib/assets/models/index.js
```diff
@@ -21,7 +21,8 @@ fs
   })
   .forEach(function(file) {
     var model = sequelize['import'](path.join(__dirname, file));
-    db[model.name] = model;
+    var modelNameCamelCase = model.name.charAt(0).toUpperCase() + model.name.slice(1)
```
sdepold: Can we call it just `modelName` ?

oss92:
yea no tests are breaking :D And I did change the variable name to `modelName`

Americas:
Should there be a test for this? This is just a convention, even if it is a good one. Should it be mandatory?

sdepold:
Tests would rock hard and its a very bad sign that nothing broke. Volunteers?

sushantdhiman:

PR #270, opened 2016-03-07
Models directory can have sub directory

IkarosCoCo:
Models directory can have sub directory

### DIFF
--- a/lib/assets/models/index.js
+++ b/lib/assets/models/index.js
@@ -24,7 +24,7 @@ function readModels(dir) {
       })
       .forEach(function (file) {
         var tmpPath = path.join(dir, file);
-        if ((fs.lstatSync(tmpPath).isDirectory() === false) && (tmpPath.slice(-3) === '.js')) {
+        if ((!fs.lstatSync(tmpPath).isDirectory()) && (tmpPath.slice(-3) === '.js')) {
           var model = sequelize['import'](tmpPath);
           db[model.name] = model;
         } else {


lib/assets/models/index.js
```diff
@@ -14,15 +14,25 @@ if (config.use_env_variable) {
...
+          var model = sequelize['import'](tmpPath);
+          db[model.name] = model;
+        } else {
+          init_db(tmpPath);
+        }
+      });
+  }
+}
+
+init_db(__dirname);
```
sdepold: Can we call it `readModels` or something like this?

lib/assets/models/index.js
```diff
@@ -14,15 +14,25 @@ if (config.use_env_variable) {
...
-  .readdirSync(__dirname)
-  .filter(function(file) {
-    return (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js');
-  })
-  .forEach(function(file) {
-    var model = sequelize['import'](path.join(__dirname, file));
-    db[model.name] = model;
-  });
+function init_db(dir) {
+  if (fs.lstatSync(dir).isDirectory()) {
```
sdepold: Can we make it a guard like this:

```
function readModels (dir) {
  if(!fs.lstatsync(dir).isDirectory())) {
    return;
  } 

  /*the rest*/
}
```

lib/assets/models/index.js
IkarosCoCo: yes,thanks for your time

lib/assets/models/index.js
IkarosCoCo: yes,It is better

lib/assets/models/index.js
```diff
@@ -14,15 +14,26 @@ if (config.use_env_variable) {
...
+  if(!fs.lstatsync(dir).isDirectory()) {
+    return;
+  }
+    fs.readdirSync(dir)
+      .filter(function (file) {
+        return file !== basename;
+      })
+      .forEach(function (file) {
+        var tmpPath = path.join(dir, file);
+        if ((!fs.lstatSync(tmpPath).isDirectory()) && (tmpPath.slice(-3) === '.js')) {
```
Americas: these conditions should be separated, you don't want it to call readModels if the file isn't ending with '.js'

You also removed the condition that does not read hidden files.

sushantdhiman:

PR #275, opened 2016-03-11
create UTC dates for migration and seed filenames

### LINKED ISSUE #274: Unnamed migrations default title lacks timezone
When generating an unnamed migration/seed filename at [lib/helpers/path-helper.js#L27](https://github.com/sequelize/cli/blob/master/lib/helpers/path-helper.js#L27) the date does not specify a timezone...

This is a problem while working with different people in different timezones. Example: 
1. Developer one creates an unnamed migration at offset -0800;
2. Shortly after, developer two creates an unnamed migration at offset +0200;
3. Even though developer one created the migration first, the migration from developer two will take place in front of migration from developer one because the date will be ahead;

One way to fix this would to just generate the dates at UTC time (or maybe set it in the options). I'd be happy to get a PR going if this is the best way. :+1:

Americas:
Fixes #274

### DIFF
--- a/lib/helpers/path-helper.js
+++ b/lib/helpers/path-helper.js
@@ -24,7 +24,7 @@ module.exports = {
   getFileName: function (type, name, options) {
     return this.addFileExtension(
       [
-        moment().format('YYYYMMDDHHmmss'),
+        moment().utc().format('YYYYMMDDHHmmss'),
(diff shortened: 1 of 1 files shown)


sdepold:
:ok_hand:

gaprl:
:+1:

PR #277, opened 2016-03-20
Prevent ugly comments highlighting

karlhorky:
This changes the highlighting language of the examples with comments from JSON to JavaScript, because JSON does not allow comments and results in the following ugly highlighting artifacts:

bildschirmfoto 2016-03-20 um 12 10 00


sdepold:
:ok_hand: I wanted to do that quite a long time already. Could you please remove the gulpfiles readme linter?

karlhorky:
Sure, done.

sdepold:
:cool:

PR #281, opened 2016-04-03
Coffeescript models skipped when loading from models dir

Mahaswami:
The index.coffee code generated to load all the models in the models directory skipping CoffeeScript models

### DIFF
--- a/lib/assets/models/index.js
+++ b/lib/assets/models/index.js
@@ -17,7 +17,7 @@ if (config.use_env_variable) {
 fs
   .readdirSync(__dirname)
   .filter(function(file) {
-    return (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js');
+    return (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js' || file.slice(-7) === '.coffee');
   })
   .forEach(function(file) {
     var model = sequelize['import'](path.join(__dirname, file));


lib/assets/models/index.js
```diff
@@ -17,7 +17,7 @@ if (config.use_env_variable) {
 fs
   .readdirSync(__dirname)
   .filter(function(file) {
-    return (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js');
+    return (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js' || file.slice(-7) === '.coffee');
   })
   .forEach(function(file) {
     var model = sequelize['import'](path.join(__dirname, file));
```
sdepold: How did we ever end up with `sequelize['import']` :scream: omg.

sdepold:
I tend to think that this change is ok :D I'm wondering if we should only load the coffee files if the coffee flag was provided during it's generation?

Mahaswami:
I intend to bring up a separate item about the coffee flag. Currently during the init time, the use of the flag is not persisted to the config file. Also the coffee flag location (as per the helper config code) is not ideal as it needed to be included under each environment as it is under the environment specific hash.

lib/assets/models/index.js
Mahaswami: Sorry I am new to sequelize and did not follow your comment. Happy to fix with further clarification

Americas:
The coffee flag can be changed to the dotfile

Mahaswami:
Sorry @Americas. I am not sure which file you are referring to. I am new to Sequelize

Americas:
:+1:

sushantdhiman:
CoffeeScript support dropped

Issue #120, opened 2015-04-07
Add --underscored option to model:create


theRemix:
by default, migrations generated from `sequelize model:create` generate camel cased timestamp fields.  

````
createdAt:
  allowNull: false
  type: DataTypes.DATE
updatedAt:
  allowNull: false
  type: DataTypes.DATE
````

add `--underscore` option to `sequelize model:create` to generate underscored column names.

````
created_at:
  allowNull: false
  type: DataTypes.DATE
updated_at:
  allowNull: false
  type: DataTypes.DATE
````

**Comment by sdepold** [comment]
Yes. I ran into that issue the other day and it totally drove me nuts :D A PR would be more than welcome otherwise it will have to wait a bit

**Comment by ojengwa** [comment]
This issues still `OPEN` even though it's added as a feature... :(

**Comment by gavinbelson** [comment]
For anyone looking, this is implemented and the --underscored parameter now works

**Comment by brunopgalvao** [comment]
What does the `--underscored` flag do when used with migrations e.g.:

```sh
npx sequelize-cli migration:generate --name rename-user_name-to-username --underscored
```

gavinbelson:
It makes the timestamp fields snake_case instead of camelCase. This usually fits with Postgres conventions better

PR #283, opened 2016-04-06
Added --underscored parameter

ichiwa:
Fixed https://github.com/sequelize/cli/issues/120

### DIFF
--- a/lib/assets/models/model.js
+++ b/lib/assets/models/model.js
@@ -12,6 +12,7 @@ module.exports = function(sequelize, DataTypes) {
         // associations can be defined here
       }
     }
+    <% if(underscored) { %>, underscored : true <% } %>
   });
 
   return <%= name %>;
--- a/lib/helpers/model-helper.js
+++ b/lib/helpers/model-helper.js
@@ -34,7 +34,8 @@ module.exports = {
   generateFileContent: function (args) {
     return helpers.template.render('models/model.js', {
       name:       args.name,
-      attributes: this.transformAttributes(args.attributes)
+      attributes: this.transformAttributes(args.attributes),
+      underscored: args.underscored ? true : false
     });
   },
 


sdepold:
Hm. There is an `underscored` parameter already and I wonder if it clashes with this implementation. Can you try to add some tests for the expected behavior?

ichiwa:
Let me explain about this implementation.
I created model with `model:create` command and `--underscored` option.
`sequelize model:create --underscored --name User --attributes first_name:string,last_name:string,bio:text`
And created following files.
```
'use strict';
module.exports = function(sequelize, DataTypes) {
  var User = sequelize.define('User', {
    first_name: DataTypes.STRING,
    last_name: DataTypes.STRING,
    bio: DataTypes.TEXT
  }, {
    classMethods: {
      associate: function(models) {
        // associations can be defined here
      }
    }
  });
  return User;
};
```
```
'use strict';
module.exports = {
  up: function(queryInterface, Sequelize) {
    return queryInterface.createTable('Users', {
      id: {
        allowNull: false,
        autoIncrement: true,
        primaryKey: true,
        type: Sequelize.INTEGER
      },
      first_name: {
        type: Sequelize.STRING
      },
      last_name: {
        type: Sequelize.STRING
      },
      bio: {
        type: Sequelize.TEXT
      },
      created_at: {
        allowNull: false,
        type: Sequelize.DATE
      },
      updated_at: {
        allowNull: false,
        type: Sequelize.DATE
      }
    });
  },
  down: function(queryInterface, Sequelize) {
    return queryInterface.dropTable('Users');
  }
};
```
Then, I was trying to make table on database with this file. (not migration command)
like this code.
```
var db = require('./models');
db.sequelize.sync({ logging:console.log });
```
but It was different from that I thought.
```
> desc Users;
+------------+--------------+------+-----+---------+----------------+
| Field      | Type         | Null | Key | Default | Extra          |
+------------+--------------+------+-----+---------+----------------+
| id         | int(11)      | NO   | PRI | NULL    | auto_increment |
| first_name | varchar(255) | YES  |     | NULL    |                |
| last_name  | varchar(255) | YES  |     | NULL    |                |
| bio        | text         | YES  |     | NULL    |                |
| createdAt  | datetime     | NO   |     | NULL    |                |
| updatedAt  | datetime     | NO   |     | NULL    |                |
+------------+--------------+------+-----+---------+----------------+
```
createdAt and updatedAt is not underscored columns so I fixed it.

Issue #247, opened 2016-01-08
Promise warnings (still got 'em)


tedbreen:
I am still getting the promise warning in **sequelize 3.17.1** when i run a migration (`sequelize db:migrate`)

#### terminal output:
```sh
Sequelize [Node: 4.2.3, CLI: 2.2.1, ORM: 3.17.1, pg: ^4.4.3]

Loaded configuration file "config/config.json".
Using environment "development".
Using gulpfile ~/movie-awards/node_modules/sequelize-cli/lib/gulpfile.js
Starting 'db:migrate'...
Finished 'db:migrate' after 54 ms
Warning: a promise was created in a handler but was not returned from it
    at ConnectionManager.getConnection (/Users/tedbreen/movie-awards/node_modules/sequelize/lib/dialects/abstract/connection-manager.js:223:23)
    at Sequelize.query (/Users/tedbreen/movie-awards/node_modules/sequelize/lib/sequelize.js:795:83)
    at QueryInterface.showAllTables (/Users/tedbreen/movie-awards/node_modules/sequelize/lib/query-interface.js:304:25)
    at ensureCurrentMetaSchema (/Users/tedbreen/movie-awards/node_modules/sequelize-cli/lib/tasks/db.js:205:6)
    at /Users/tedbreen/movie-awards/node_modules/sequelize-cli/lib/tasks/db.js:24:9
    at /Users/tedbreen/movie-awards/node_modules/sequelize-cli/lib/tasks/db.js:321:7
    at processImmediate [as _immediateCallback] (timers.js:383:17)
From previous event:
    at Promise.then (/Users/tedbreen/movie-awards/node_modules/sequelize/lib/promise.js:21:17)
    at getMigrator (/Users/tedbreen/movie-awards/node_modules/sequelize-cli/lib/tasks/db.js:320:6)
    at Object.module.exports.db:migrate.task (/Users/tedbreen/movie-awards/node_modules/sequelize-cli/lib/tasks/db.js:23:7)
    at Gulp.gulp.task.aliases (/Users/tedbreen/movie-awards/node_modules/sequelize-cli/lib/helpers/gulp-helper.js:14:14)
    at module.exports (/Users/tedbreen/movie-awards/node_modules/orchestrator/lib/runTask.js:34:7)
    at Gulp.Orchestrator._runTask (/Users/tedbreen/movie-awards/node_modules/orchestrator/index.js:273:3)
[...]

PR #289, opened 2016-04-24
fixes Bluebird promise warnings

leebenson:
- fixes sequelize/cli/issues/242
- fixes sequelize/cli/issues/247

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -21,7 +21,7 @@ module.exports = {
     },
 
     task: function () {
-      getMigrator('migration', function (migrator) {
+      return getMigrator('migration').then(function (migrator) {
         return ensureCurrentMetaSchema(migrator).then(function () {
           return migrator.pending();
         }).then(function (migrations) {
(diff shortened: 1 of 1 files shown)


sdepold:
👌

Americas:
it fails to fix the warnings, as I said before this issue is not cli related.

But I like this change, promises > callbacks :laughing:

leebenson:
@Americas it fixed them on my end.

Americas:
ahah, I linked to the wrong cli package. My bad!

It does work. But it is strange, since there was no unreturned promise to begin with :laughing:

leebenson:
The issue starts with the gulp task. As @opatut mentioned in #247, running `db:migrate` yielded something close to this:

> Finished 'db:migrate' after 99 ms

... yet the task would continue to run after gulp considered it finished. By returning the `sequelize` chain inside the task and transforming the called `getMigrator` function to use promises, there's now a neat chain for gulp to manage the task.

Bottom line: Always return inside a promise.

Americas:
I'm not questioning the fact that gulp returned before it finished running. Granted you can fix that and at the same time fix the promise warnings, you should also be able to fix the promise warnings without fixing the gulp early finish. And that was what was bothering, but I finally figured it out! saul goodman! :laughing:

PR #379, opened 2016-10-29
Add db:migrate:pending to check for pending migrations, closes #165

### LINKED ISSUE #165: Possible to check for pending migrations on startup?
This isn't really an issue per se but I didn't see another forum for it.

Is it possible to configure Sequelize to check for pending migrations on startup? We've had numerous cases where devs forget to run migrations and proceed to chase their tails trying to figure out why their code suddenly stopped working.

golmansax:
Threw together a pull to tackle #165, got tests to pass locally.  A good amount of the src and test code was copied from `db:migrate`, so definitely open to feedback!

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -63,7 +63,7 @@ module.exports = {
             migrations.forEach(function (migration) {
               console.log(migration.file);
             });
-            process.exit(1);
+            process.exit(3);
           }
         }).catch(function (err) {
           console.error(err);
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -41,6 +41,38 @@ module.exports = {
     }
   },
 
+  'db:migrate:pending': {
+    descriptions: {
+      'short': 'List pending migrations.',
+      'long': [
+        'The command lists pending migrations and exits with non-zero status if',
+        'there are any.'
+      ]
+    },
+
+    task: function () {
+      return getMigrator('migration').then(function (migrator) {
+        return ensureCurrentMetaSchema(migrator).then(function () {
+          return migrator.pending();
+        }).then(function (migrations) {
+          if (migrations.length === 0) {
+            console.log('No pending migrations, database schema up to date.');
+            process.exit(0);
+          } else {
+            console.log('Pending migrations:');
+            migrations.forEach(function (migration) {
+              console.log(migration.file);
+            });
+            process.exit(1);
+          }
+        }).catch(function (err) {
+          console.error(err);
+          process.exit(1);
+        });
+      });
+    }
+  },
+
   'db:seed': {
     descriptions: {
       'short': 'Run specified seeder.',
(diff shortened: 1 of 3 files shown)


lib/tasks/db.js
```diff
@@ -41,6 +41,38 @@ module.exports = {
...
+          } else {
+            console.log('Pending migrations:');
+            migrations.forEach(function (migration) {
+              console.log(migration.file);
+            });
+            process.exit(1);
+          }
+        }).catch(function (err) {
+          console.error(err);
+          process.exit(1);
```
cusspvz: `process.exit(255);`?
I wouldn't exit with the same exit code as if it were pending migrations.

test/db/migrate/pending.test.js
```diff
@@ -0,0 +1,87 @@
...
+          callback(null, result);
+        });
+    };
+
+    it('fails with a not 0 exit code and reports pending migrations', function (done) {
+      prepare(function (_, stdout) {
+        expect(stdout).to.contain('createPerson.js');
+        done();
+      }, {
+        cli: { exitCode: 1 },
```
cusspvz: If there is an error, this test will pass because errors are exiting with the same exit code as "pending migrations".

golmansax:
@cusspvz Thanks for the fast reply!  It looks like an exit code of 1 is used throughout the repo, so I instead changed the pending migration exit code to 3 (wasn't sure what exit code to choose, and 2 is reserved according to http://www.tldp.org/LDP/abs/html/exitcodes.html)

You can see the changes in 69391de.  Thoughts?

cusspvz:
I've took a time to look the repo and seems it [only exits with 1](https://github.com/sequelize/cli/blob/cba03c8f17740cd652c40103830d4cff9e44f86b/lib/tasks/db.js#L28-L39) when it has errors, and with `0` when it hasn't, even in situations you want to express nothing has happened. In fact, it makes some sense, otherwise command would break logic chains (like `sequelize-cli db:migrate:pending && echo "OK"`).

Please note that I'm only an enthusiast/contributor, @sdepold is the admin of this repo.

golmansax:
If that's the case, then maybe it can always exit 0 by default, and then running `db:migrate:pending --fail` or with some other flag will cause it to exit 1 when then are pending migrations..  Thoughts?

Americas:
I like this functionality, but the name is probably not the best. #385 has a better idea, and what I would propose is merging this change in that one.

What I propose:

- Create the command `db:migrate:status` as is in #385 
- Add 2 optional arguments to that command:
- - `--pending` outputs only pending migrations
- - `--executed` outputs only executed migrations


Also, about the exit code, I believe it best to add the `--fail` flag to tell the program to exit with error on pending migrations. I assume this will be used to programatically chain functionality to the end of this call, it is better to add this as a special use case.

golmansax:
Yep, that makes sense.  Maybe we can put this pull on hold until after #385 is merged, and then I'll update my pull to add onto the `db:migrate:status` command.

Americas:
for me that would be perfect :+1:

sushantdhiman:

PR #385, opened 2016-11-01
add db:migrate:status

gotrevor:
This also addresses #165 , with a slightly different approach from #379.

Both `db:migrate:status` and `db:migrate:pending` would probably be useful.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -165,6 +165,35 @@ module.exports = {
     }
   },
 
+  'db:migrate:status': {
+    descriptions: {
+      'short': 'List the status of all migrations'
+    },
+
+    task: function () {
+      return getMigrator('migration').then(function (migrator) {
+        return ensureCurrentMetaSchema(migrator).then(function () {
+          return migrator.executed();
+        }).then(function (migrations) {
+          _.forEach(migrations, function (migration) {
+            console.log('up  ', migration.file);
+          });
+        }).then(function () {
+          return migrator.pending();
+        }).then(function (migrations) {
+          _.forEach(migrations, function (migration) {
+            console.log('down', migration.file);
+          });
+        }).then(function () {
+          process.exit(0);
+        }).catch(function (err) {
+          console.error(err);
+          process.exit(1);
+        });
+      });
+    }
+  },
+
   'db:migrate:undo': {
     descriptions: {
       'short': 'Revert the last migration run.'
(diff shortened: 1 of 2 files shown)


lib/tasks/db.js
```diff
@@ -165,6 +165,35 @@ module.exports = {
...
+    descriptions: {
+      'short': 'List the status of all migrations'
+    },
+
+    task: function () {
+      return getMigrator('migration').then(function (migrator) {
+        return ensureCurrentMetaSchema(migrator).then(function () {
+          return migrator.executed();
+        }).then(function (migrations) {
+          _.forEach(migrations, function (migration) {
```
Americas: I don't like the way you are logging the results. Maybe something like this would be nicer:
```
Executed migrations:
- <migration file 1>
- <migration file 2>
...

Pending migrations:
- <migration file 3>
- <migration file 4>
...
```

If there are no executed or pending migrations, instead of that just log:

```
No executed migrations found.

No pending migrations found.
```

Americas:
I have written this in #379, I believe these two commands should exist as only one command, with optional arguments to tweak the result. WDYT of merging the functionalities of #379 with this command?

gotrevor:
@Americas  Either one of these PRs will be fine with me.  I'm mostly interested in seeing down migrations.

gotrevor:
ActiveRecord has `db:migrate:status`, very similar to this PR.
It has another command `db:migrate: abort_if_pending_migrations`, which is very similar to #379 :
https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/railties/databases.rake#L165

Following their precedent of two separate commands, instead of one with options makes sense to me.  It also makes for a nice parallel between the two migrations systems, which seems valuable.

lib/tasks/db.js
Americas: I see your point here, it is helpful for chaining more commands on the output.

Americas:
@sdepold @cusspvz opinions?

Should we create 2 commands or just add options to this one so it errors if there are pending migrations?

The rest LGTM

gotrevor:
@Americas @sdepold @cusspvz
Based on this comment:
https://github.com/sequelize/cli/pull/379#issuecomment-258137064
I think that this PR is ready to merge, and that one will be re-worked on top of it.

The question of "Should we create 2 commands or just add options to this one so it errors if there are pending migrations?" can be handled over there.

Can this get merged?

Thanks,
Trevor

PR #393, opened 2016-11-28
Include error details when seeder throws an exception

broofa:
Please don't drop important information like line number and stack trace on the ground when reporting errors.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -67,7 +67,7 @@ module.exports = {
           process.exit(0);
         })
         .catch(function (err) {
-          console.error('Seed file failed with error:', err.message);
+          console.error('Seed file failed with error:', err);
           process.exit(1);
         });
       });
@@ -96,7 +96,7 @@ module.exports = {
           process.exit(0);
         })
         .catch(function (err) {
-          console.error('Seed file failed with error:', err.message);
+          console.error('Seed file failed with error:', err);
           process.exit(1);
         });
       });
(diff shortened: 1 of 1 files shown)


cusspvz:
Most of the errors could be linted, but indeed, I think the stack could be outputed. In case @sdepold  does not agree, another solution would be a stack-show feature based on a passed arg or environment variable.

broofa:
Linted?  That works for some subset of JS errors, but certainly not for all.  Maybe I misunderstand you though. Either way it's just making life harder than necessary if the info needed to diagnose an error is there and just not being logged.

(Note: Even with this pull request there are cases where important information may be obscured.  For example, if there's a validation error, it would be helpful to surface the `err.errors` information.)

cusspvz:
> That works for some subset of JS errors, but certainly not for all.

Yep, I know, thats why it should be landed, or at least, give the user that choice. We'll have to wait for @sdepold.

>  Maybe I misunderstand you though.

Probably, what I meant was, that on development environment, most error-cases are syntax type or typos they are easy to avoid/diagnose, but of course I agree with you on that it is a must to show up the entire error stack. Also there is a scenario where a migration or seed could go wrong on a staging-like environment for any random reason, and this would definitely aid to diagnose and iterate faster.

> (Note: Even with this pull request there are cases where important information may be obscured. For example, if there's a validation error, it would be helpful to surface the err.errors information.)

I think it wouldn't be a bad idea to introduce an util method to handle errors and couple it with async tasks, or even wrap it around the methods on the [gulpfile](https://github.com/sequelize/cli/blob/3fef9d8bb7379293186fff0edb0fa86db3b83ab4/lib/gulpfile.js#L23).

sushantdhiman:
Fixed in https://github.com/sequelize/cli/pull/411

PR #404, opened 2016-12-20
Fix small markup typo

jonatasrancan:
Fixing small typo on README


sushantdhiman:
I think docs are referring to node imports i.e. `require()`, hence the verb `require`d. You can change it something like "file will get imported"

Americas:
I prefer the way it is. It explicitly says that the `.sequelizerc` file is passed to `require`, so that file can be anything that `require` can handle.

jonatasrancan:
hum.....
So after all, it's not a typo, I just got it wrong.

jonatasrancan:
@Americas going to close then. 
ok?

PR #430, opened 2017-03-02
Move gulp to devDependencies

blacksonic:
Gulp is only used for testing, it is not necessary to be included in npm package installation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,6 @@
     "cli-color": "~0.3.2",
     "findup-sync": "^0.4.0",
     "fs-extra": "^1.0.0",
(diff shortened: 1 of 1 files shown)


sushantdhiman:
CLI is based on gulp https://github.com/sequelize/cli/blob/master/lib/gulpfile.js#L3

blacksonic:
@sushantdhiman  But it is NOT used in the command line tool https://github.com/sequelize/cli/blob/master/bin/sequelize
It is trivial that gulp is used in a gulp file...

The gulp file is only used for testing.

PR #440, opened 2017-03-22
[FIX #439] return the promise

### LINKED ISSUE #439: Promise warning in gulp-helper
I am getting the following warning from the code in gulp-helper when I load a config via a promise. 

Warning: a promise was created in a handler at /node_modules/sequelize-cli/lib/helpers/gulp-helper.js:17:16 but was not returned from it. I think the issue is a missing return on the line indicated. https://github.com/sequelize/cli/blob/master/lib/helpers/gulp-helper.js#L17

gotrevor:


gotrevor:
Thx, @mention-bot Looks like you got a good list of potential reviewers.  Perhaps, also, @sushantdhiman

PR #447, opened 2017-04-14
Improve documentation on ssl config for the CLI

zamiang:
Somewhat inspired by this issue: https://github.com/sequelize/cli/issues/154 where the last comment has 22 thumbs up at time of posting this issue. Their config is out of date but the activity there may be an indication that the documentation is lacking. Added an explicit section on SSL to clear that up

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,22 @@ module.exports = {
 }
 ```
 
+### Configuration for connecting over SSL
+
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Fixed https://github.com/sequelize/cli/tree/master/docs#configuration-for-connecting-over-ssl

PR #459, opened 2017-05-15
Add option to db:migrate:undo:all which allows reverting migration to a specific version.

kristijanhusak:
Add option to undo migrations to the specific version.
https://github.com/sequelize/umzug#reverting-executed-migration

This option is very useful when continuous integration is used, since it allows easier rolling back in case something went wrong.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -231,7 +231,10 @@ module.exports = {
 
   'db:migrate:undo:all': {
     descriptions: {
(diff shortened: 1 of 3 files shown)


kristijanhusak:
@Americas I added the test. I haven't run the whole suite, just that single one, since it is taking too long to finish.

Americas:
`test/db/migrate/undo/all_to.test.js: line 48, col 115, Line is too long.`

PR #461, opened 2017-05-16
Improve sequelize options handling

Americas:
We are already parsing config file options in `config-helper.js` and we should keep this all in the same place. This PR moves the `use_env_variable` to `config-helper.js` and removes all the extra logic from `tasks/db.js`.

Also fixed the database and ssl parameters, so an url string with `...?ssl=true` does not break anymore.

### DIFF
--- a/lib/helpers/config-helper.js
+++ b/lib/helpers/config-helper.js
@@ -128,17 +128,22 @@ var api = {
       if (api.rawConfig[env]) {
         console.log('Using environment "' + env + '".');
 
-        // The Sequelize library needs a function passed in to its logging option
-        if (api.rawConfig.logging && !_.isFunction(api.rawConfig.logging)) {
-          api.rawConfig.logging = console.log;
-        }
-
         api.rawConfig = api.rawConfig[env];
+      }
+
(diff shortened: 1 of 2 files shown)


lib/helpers/config-helper.js
```diff
@@ -176,10 +178,11 @@ var api = {
     try {
       var urlParts = url.parse(urlString);
       var result   = {
-        database: urlParts.path.replace(/^\//,  ''),
+        database: urlParts.pathname.replace(/^\//,  ''),
```
sushantdhiman: likely it wont work with `sqlite:memory` url but we dont need cli / migration for in memory setups.

Americas: @sushantdhiman [this](https://github.com/sequelize/cli/pull/461/files#diff-072bd6704a4a833da23b826bb978aaf7R208) is the fix you added in sequelize, should work now right?

sushantdhiman: yeah it should now

Americas:
These tests are dangerous, they were failing but turning up green...

Fixed them now, and I think this can be merged when the tests complete.

PR #479, opened 2017-07-03
Fix Relative Path on init for Windows Users

### LINKED ISSUE #478: `sequelize init` Incorrectly writes path to config.json
After running `sequelize init`, the index.js file in the models folder specifies an invalid path to config.json due to the back-slashes being read as escape characters.

```javascript
var config    = require(__dirname + '/..\config\config.json')[env];
```

This throws an error:
```
Error: Cannot find module '<path_to_directory>\models/..configconfig.json'
```

**Should be**
```javascript
var config    = require(path.join(__dirname + './../config/config.json'))[env];
```

I'm happy to contribute, but the tests won't run for me locally. I'm on Windows 10/powershell. I also tried the built in bash terminal that Windows has now, but no dice.  I get an error `<path-to-binary>\sequelize is not recognized as an internal or external command` from the tests. Works fine when I run `sequelize` on my own though.

_Edit: Simplified_

djbreen7:
Resolves #478 

The test in init.test.js already checks for forward-slashes, so no tests
need to be updated.

### DIFF
--- a/lib/helpers/init-helper.js
+++ b/lib/helpers/init-helper.js
@@ -73,7 +73,7 @@ module.exports = {
       helpers.asset.write(
         indexPath,
         helpers.template.render('models/index.js', {
-          configFile: '__dirname + \'/' + relativeConfigPath + '\''
+          configFile: '__dirname + \'/' + relativeConfigPath.replace(/\\/g, '/') + '\''
         }, {
           beautify: false
         })


djbreen7:
I have a feeling that reviewers will suggest using `path.join`. I did try that, but it uses backslashes for Windows which become escape characters.

PR #484, opened 2017-07-16
Update Readme.md

leonardo0543:
Adding a cli model generation example


sushantdhiman:
I have talked about this in other PR as well https://github.com/sequelize/cli/pull/487#issuecomment-317256099

These commands are listed in Readme already, but if you think its hard to find them I will prefer a FAQ.md over adding these instruction in Readme file itself

PR #487, opened 2017-07-20
migration schema

ralvarezo123:
Command to execute and return all migrations


sushantdhiman:
Already listed in https://github.com/sequelize/cli/blob/master/README.md#global-install-usage , what is your motivation behind this PR.

If you want to include it may be FAQ.md will be suited for this

Issue #482, opened 2017-07-10
Update the Skeleton


GuillaumeLeclerc:
Hello,

Since ES6 is now supported in the LTS version of node for a while (except imports). I think it might be a good idea to update the skeletons generated by the CLI. Or give the ability to the user to override them.

Any thoughts ?

Jadex1:
I wold like to see the this too. Also this is my first time commenting on an open source project. I'm not sure if I'm suppose to just like this or comment it to get attention. I just started using sequelizer and my project is in es6.

Issue #389, opened 2016-11-13
es2015?


evgenyboxer:
I'm using es2015 in my project, and I just started using the CLI.
How can I use it in es2015 env?

gotrevor:
That should not be a problem.  What problems are you having?

PR #493, opened 2017-07-27
Adds the argument --es6 in the command migration:create and seed:create

felpin:
The purpose of the PR is to add the option `--es6` to the command `migration:create`.

When executing `migration:create --es6` it will use a skeleton in ES6 format with arrow functions.

This PR resolves issue #482, although it is not a breaking change required for major version. Maybe a future PR can change the default skeleton to be in ES6 format.

This PR may resolve issue #389 if it is a duplicate from #482.

OBS: When executing `migration:create --es6 --coffee`, the argument `--es6` will be ignored. The reason is that the js file would be converted to a coffee file. 'Would' is the correct word here because [js2coffee](https://www.npmjs.com/package/js2coffee) doesn't support ES6 yet (See [issue](https://github.com/js2coffee/js2coffee/issues/449))

### DIFF
--- a/lib/assets/migrations/skeleton-es6.js
+++ b/lib/assets/migrations/skeleton-es6.js
@@ -1,0 +1,21 @@
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    /*
+      Add altering commands here.
+      Return a promise to correctly handle asynchronicity.
+
+      Example:
+      return queryInterface.createTable('users', { id: Sequelize.INTEGER });
+    */
+  },
+
+  down: (queryInterface, Sequelize) => {
+    /*
+      Add reverting commands here.
+      Return a promise to correctly handle asynchronicity.
+
+      Example:
+      return queryInterface.dropTable('users');
+    */
+  }
+};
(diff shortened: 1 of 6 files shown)


felpin:
I've just realized that there is another skeleton for `seed:create`.

Well... I'll work on this to support `--es6` argument as well.

Jadex1:
What does OBS stand for?

felpin:
Observation

Where I live, it's a common word to use when you want to write something that the reader must pay attention. I didn't know it wasn't usual outside my country.

Jadex1:
Ah, I was trying not to assume anything. What country are you from? I’m also new to Open Source and I didn’t know if it was something specific to the project or not.

- James

felpin:
Does anyone know how to run CI again? The check has failed because
`The job exceeded the maximum time limit for jobs, and has been terminated.`

sushantdhiman:
Fixed in https://github.com/sequelize/cli/commit/733ced9e55b48d7aa87cfb964e37a028bc5c95d4

PR #470, opened 2017-06-16
moving model.associate outside classMethods due to sequelize@4.0.0

geluso:
I'm trying to figure out if sequelize-cli is supposed to match breaking changes introduced in sequelize@4.0.0. If yes, the model.associate method needs to be moved outside `classMethods` because `classMethods` was removed in sequelize@4.0.0.

Review my comment here: https://github.com/sequelize/cli/pull/468

PR #499, opened 2017-08-01
changed model template  - assosciations compatible with current sequelize

deejes:
Sequelize takes model associations in an updated format. Edits to the template now make it compatible with the current version of Sequelize.


lib/assets/models/model.js
```diff
@@ -6,14 +6,11 @@ module.exports = function(sequelize, DataTypes) {
       <%= fieldName %>: DataTypes.<%= attributes[fieldName].toUpperCase() %>
       <%= (Object.keys(attributes).length - 1) > index ? ',' : '' %>
     <% }) %>
-  }, {
-    <%= underscored ? 'underscored: true,' : '' %>
```
eseliger: What happened to this option?

deejes: Removed it by mistake. Adding it back.

sushantdhiman:
https://github.com/sequelize/cli/pull/470

PR #503, opened 2017-08-04
Update FAQ.md

salvarezsc:
New commands


sushantdhiman:
These are very specific commands already explained in Readme doc. I want to keep FAQ lean for now, just dont want to add every other command available to it

PR #506, opened 2017-08-05
Update FAQ.md

Emanuel94:
Add new commando in FAQ.md


sushantdhiman:
I dont want to add every other command to FAQ, I think someone migrating to old schema is already an avid user of CLI or at least have read Readme docs.

Issue #70, opened 2014-12-08
Add db:create and db:drop tasks


cesarandreu:
I come from a Rails background and think it'd be pretty convenient to have `sequelize db:create` and `sequelize db:drop` as part of the cli.

Since I only use Postgres I wrote two scripts to achieve this and published it under [sequelize-migration-pg-extras](https://github.com/cesarandreu/sequelize-migration-pg-extras).

If you could give some advice on how to go about adding native support, I'd be willing to try writing a PR.

th3m4ri0:
+1

sladebot:
+1

sdepold:
PRs are welcome :)

sadaf2605:
+1

jackycute:
+1

HOuaghad:
:+1:

joaodiogocosta:
+1

juniorplenty:
+1

joost:
+1

Catchpowle:
+1

anthony-ism:
+1

th3m4ri0:
@joost @Catchpowle @anthony-ism  
Since March 10, 2016, it is possible to [use emoji reactions on comments](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments). Please use the 👍 reaction to show your interest instead of commenting "+1".  This saves everyone in the thread a notification. If you want to get updates on the issue in the future, you can use the "Subscribe" button. Thank you ! 😄

StlthyLee:
Has anything been done on adding these? At present this is the only thing(s) stopping me from using sequelize in my companies project.

quintencls-4c:
@StlthyLee  I just started using Sequelize and haven't found any news on this issue. Although it is quite easy to add the missing db tasks (db:create & db:drop) yourself with [pg lib](https://www.npmjs.com/package/pg). I defined the folowing scripts in my package.json for development:
```
"scripts": {
    "start": "node index.js",
    "serve": "nodemon index.js",
    "db:create": "node ./db/tasks/create_local_db.js",
    "db:drop": "node ./db/tasks/drop_local_db.js",
    "db:migrate": "sequelize db:migrate --env development",
    "db:migrate:undo": "sequelize db:migrate:undo --env development",
    "db:migrate:undo:all": "sequelize db:migrate:undo:all --env development",
    "db:recreate": "npm run db:drop & npm run db:create & npm run db:migrate",
[...]

PR #510, opened 2017-08-09
Add db:create and db:drop for MySQL/Postgres 🎊

aledalgrande:
Needed for a better dev experience. I didn't use `sequelize` because it doesn't have support for opening a connection without a database, but the dialect libraries work.

Feel free to request any changes.

Solves #70.

### DIFF
--- a/lib/tasks/db.js
+++ b/lib/tasks/db.js
@@ -11,6 +11,34 @@ var Umzug     = require('umzug');
 var clc       = require('cli-color');
 
 module.exports = {
+  'db:create': {
+    descriptions: {
+      'short': 'Creates the configured database.',
+      'long': [
+        'The command creates the configured database.'
+      ]
+    },
+
+    task: function () {
+      var config = helpers.config.readConfig();
+      var database = config.database;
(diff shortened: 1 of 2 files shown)


dagumak:
I was surprised to find that this was missing!

dagumak:
@eseliger Thoughts?

dagumak:
@aledalgrande 
```
$ sequelize db:create

Sequelize [Node: 7.7.3, CLI: 2.8.0, ORM: 4.7.5]

Loaded configuration file "config/config.json".
{ Error: Cannot find module 'pg'
```

lib/tasks/db.js
```diff
@@ -437,6 +465,59 @@ function getMigrator (type) {
...
+          process.exit(0);
+        }
+      });
+    }
+    else if (dialect.match('postgres')) {
+      var Postgres = require('pg').Client;
+      var client = new Postgres({
+        user: config.username,
+        host: config.host,
+        database: 'postgres',
```
dagumak: `config.database`?

aledalgrande: @dagumak nope, I am passing `postgres` because the database doesn't exist yet, so I am using the default one.

PR #512, opened 2017-08-12
fix error of creating config directory recursively.

everthis:
Current `mkdirSync` creates a directory successfully only if the parent directory exists.
Error arises in the following example:

create a  `.sequelizerc` file in project root directory, for example:
```js
var path = require('path');

module.exports = {
  'config': path.resolve(__dirname, './db/config/config.js'),
  'migrations-path': path.resolve(__dirname, './db/migrations'),
  'models-path': path.resolve(__dirname, './db/models'),
  'seeders-path': path.resolve(__dirname, './db/seeders')
}
```
execute `./node_modules/.bin/sequelize init`

se

### DIFF
--- a/lib/helpers/config-helper.js
+++ b/lib/helpers/config-helper.js
@@ -94,7 +94,7 @@ var api = {
     var configPath = path.dirname(api.getConfigFile());
 
     if (!helpers.path.existsSync(configPath)) {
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Fixed by https://github.com/sequelize/cli/pull/539

PR #522, opened 2017-08-23
Fix readme

YUFENGWANG:
add extra comment field in json example file, fix README.md display error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,13 +214,13 @@ database, using `sequelize`, but want to use a different table, you can change t
     "host": "127.0.0.1",
     "dialect": "mysql",
 
-    // Use a different storage type. Default: sequelize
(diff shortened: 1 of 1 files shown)


eseliger:
Why is this necessary though? 
Whats the improvement of having invalid JSON because the same key was used multiple times instead of a comment? The comment is way easier to identify, I think.

YUFENGWANG:
Yup, comment is way easier  to identify, but comment is not allowd in json. It will cause markdown display issue. Look at the red spot in README.md. I just feel uncomfortable seeing this.[here is the discussion](https://stackoverflow.com/questions/244777/can-comments-be-used-in-json). Feel free to ignore it
markdown
 if you don't like this way.

sushantdhiman:
This way to refer comments as other fields is very flaky, I am ok with what we have currently

Issue #545, opened 2017-09-19
The db:create command fails if the database name contains a hyphen


aledalgrande:
## What you are doing?
The config file:
```
{
  "development": {
     "username": "postgres",
     "password": "",
     "database": "example-development",
     "dialect": "postgres",
     "host": "localhost"
  }
}
```

The command:
```
sequelize db:create
```

## What do you expect to happen?
The database is created.

## What is actually happening?
The code at https://github.com/sequelize/cli/blob/master/src/commands/database.js#L22 should be using quotes to make this work I think.

_Output, either JSON or SQL_

```
Error: syntax error at or near "-"
```

__Dialect:__ postgres
__Database version:__ 9.6.3
__Sequelize CLI version:__ 3.0.0-3
__Sequelize version:__ 4.8.4

Xanderite:
I actually just came across this bug, and have been thinking about a solution, I suppose we could wrap all database names with ""'s this definitely fix the issue in this situation, not sure what other effects it will have.
Would a quick fix in this case be to just add the quotes in the json definition i.e.:
```
{
  "development": {
     "username": "postgres",
     "password": "",
     "database": "\"example-development\"",
     "dialect": "postgres",
     "host": "localhost"
  }
}
```
I'm trying to use this tool with the `--url` connection strings as we have many different testing environments that frequently change, its a bit of a pain to manage them in json, so because all the dbs have been created with hyphens in their name, this presents a problem as adding quotes to the string:
```
'postgres://user:password@host:port/database-name'
```
Will fail, but you can't add escaped quotes because they are uri encoded by urs.parse()
```
'postgres://user:password@host:port/"database-name"'
```
will create a config with database name of: "%22database-name%22"

I've created a fork of the project to decodeURI() the dbname to allow this, not sure if this is the best approach though. Maybe all postgres dbnames should be wrapped with quotes instead?

Xanderite:
Hey, this is still an issue, is it ok if we re-open it? I have made a PR that has tests and should fix the issue. Which is also a big problem for connection string usage, as you cannot wrap the database name with quotes as double-quotes get encoded by url.parse to `%22`.

Xanderite:
@sushantdhiman

Xanderite:
I misread this thread, the issue has been fixed in 3.0.0.

PR #555, opened 2017-09-23
Bug/postgres hyphenated db names

Xanderite:
Addresses issues in https://github.com/sequelize/cli/issues/545
- Wrap database name with quotes in create and drop scripts, if dialect is Postgres.
- Update tests to use hyphens in Postgres create and drop db tests, to prevent regression.

### DIFF
--- a/src/commands/database.js
+++ b/src/commands/database.js
@@ -17,6 +17,12 @@ exports.handler = async function (args) {
   const sequelize = getDatabaseLessSequelize();
   const config = helpers.config.readConfig();
 
+  switch (config.dialect) {
+    case 'postgres':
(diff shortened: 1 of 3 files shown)


sushantdhiman:
I have fixed this by using sequelize quoteIdentifier , and even tests are passing (in master) https://github.com/sequelize/cli/blob/f4090948eee793d43fd758f81384f01480de0643/test/db/db-create.test.js#L45

Are you certain you are trying this on v3.0.0 ?

Xanderite:
Aha, I must admit I did not check to see if it had been fixed since I made a quick hacky solution on Thursday. Did not see why the ticket was closed, my bad.

PR #562, opened 2017-09-29
model.js

vleitt:
Adjust to associations definition according to changes in version 4 of sequelize


sushantdhiman:
https://github.com/sequelize/cli/pull/470

PR #576, opened 2017-10-23
Fix #222 - stop parsing attributes and exit if type is unknown

### LINKED ISSUE #222: Validators for Data Types
You should not be able to enter data types that do not exist in Sequelize. 

```
node_modules/.bin/sequelize model:create --name Todo --attributes "name:RYANGOSLING"
```

Results in:

```
'use strict';
module.exports = {
  up: function(queryInterface, Sequelize) {
    return queryInterface.createTable('Todos', {
      id: {
        allowNull: false,
        autoIncrement: true,
        primaryKey: true,
        type: Sequelize.INTEGER
      },
      name: {
        type: Sequelize.RYANGOSLING
      },
      createdAt: {
        allowNull: false,
        type: Sequelize.DATE
      },
      updatedAt: {
        allowNull: false,
        type: Sequelize.DATE
      }
    });
  },
  down: function(queryInterface, Sequelize) {
    return queryInterface.dropTable('Todos');
  }
};
```

jraut:


src/helpers/model-helper.js
```diff
@@ -2,19 +2,41 @@ import helpers from './index';
...
     set.forEach(pair => {
       const split = pair.split(':');
-
+    
+      if (knownDataTypes.indexOf(split[1].toUpperCase()) === -1) {
+        const errorText = dialect
+          ? 'Unknown attribute type ' + split[1] + ' not supported in ' + dialect
+          : 'Unknown attribute type ' + split[1] + ' not supported in Sequelize';
+        
+        console.error(errorText);
```
sushantdhiman: Use `helpers.view.error` method instead, no need to exit with error code with that

src/helpers/model-helper.js
```diff
@@ -2,19 +2,41 @@ import helpers from './index';
...
+    /*
+     * SpecialDataTypes is used to filter out values postgres, mysql, sqlite, mssql.
+     * It should be read from postgres instead of being enumerated.
+     */
+    const specialDataTypes = ['postgres', 'mysql', 'sqlite', 'mssql'];
+
+    /*
+     * Use either dataTypes for the dialect or dataTypes from Sequelize
+     */
+    const knownDataTypes = dialect
```
sushantdhiman: As dialect name is always defined other condition will never execute?

jraut: Sure. The extra should be removed then. updating.

src/helpers/model-helper.js
jraut: Sure!

sushantdhiman:
@jraut Any update :)

pranavpr:
Can I work on this if no one is working on it?

PR #579, opened 2017-10-25
Improved database defaults

dijonkitchen:
Uses `process.env` variables to control config for production (as [documented](http://docs.sequelizejs.com/manual/tutorial/migrations.html)).

Details in commits.

### DIFF
--- a/src/helpers/config-helper.js
+++ b/src/helpers/config-helper.js
@@ -81,10 +81,10 @@ const api = {
         dialect: 'mysql'
       },
       production: {
-        username: 'root',
-        password: null,
-        database: 'database_production',
-        host: '127.0.0.1',
+        username: process.env.PROD_DB_USERNAME,
+        password: process.env.PROD_DB_PASSWORD,
+        database: process.env.PROD_DB_NAME,
+        host: process.env.PROD_DB_HOSTNAME,
         dialect: 'mysql'
       }
     }, undefined, 2) + '\n';


src/helpers/config-helper.js
```diff
@@ -69,22 +69,22 @@ const api = {
       development: {
         username: 'root',
         password: null,
-        database: 'database_development',
+        database: `${process.env.PWD.split('/').pop()}_development`,
```
sushantdhiman: Remove this change

src/helpers/config-helper.js
```diff
@@ -69,22 +69,22 @@ const api = {
...
         host: '127.0.0.1',
         dialect: 'mysql'
       },
       test: {
         username: 'root',
         password: null,
-        database: 'database_test',
+        database: `${process.env.PWD.split('/').pop()}_test`,
```
sushantdhiman: Remove this one as well

sushantdhiman:
Should have added tests, this config actually doesnt works. Default config is missing "production" keys which use env variable

dijonkitchen:
Should there be tests for a production environment? I figure that'd be abstracted away by server environment variables per [12factor principles](https://12factor.net/config).

sushantdhiman:
Issue is if those env variable are not defined, `config.json` will not include those keys. We should definitely have tests if correct default config is generated, i.e. env defaults are actually added in config file rather than empty / non existing keys

sushantdhiman:
I am reverting this change, documentation should specify safe defaults, beside having env var inside json is not good either. There is a proper way of doing that as explained in http://docs.sequelizejs.com/manual/tutorial/migrations.html#using-environment-variables

Issue #155, opened 2015-07-16
Specify array type with model:create


sdroadie:
As far as I can tell, there is no way to assign types to arrays using model:create. The syntax could be pretty simple: `sequelize model create --name Model --attributes 'info:array:string'`. Currently, this will pass, but just create an attribute that is an Array type. Without a type specified for the array, however, `db:migrate` fails.

Sorry if there is already a means for doing this: if so, I didn't see it after perusing the codebase.

sdepold:
Yeah you are right. There is currently no way of doing this. Want to give it a try?

link-alex:
Looks like this issue is still valid so I gave it a try :blush:

link-alex:
Ok, so I forgot that String.prototype.includes appeared in node v6, so will update to support node v4+

sushantdhiman:
Added by https://github.com/sequelize/cli/pull/589, Thanks @link-alex

CognitiveClouds-Prasad:
So, how can I use this in sequelize cli? I am not able to migrate. There is no documentation I could find after about 3 hours of search.

ccemeraldeyes:
> There is no documentation I could find after about 3 hours of search.

yeah guys the docs for the cli are impossible to find pls fix

JoeRoddy:
For anyone stuck on this, the syntax is like:
`npx sequelize-cli model:generate --name Player --attributes favoriteColors:array:string`

^docs would be great for this!

PR #589, opened 2017-11-15
Feature: specify array type with model:create

link-alex:
for https://github.com/sequelize/cli/issues/155

### DIFF
--- a/src/assets/migrations/create-table.js
+++ b/src/assets/migrations/create-table.js
@@ -11,9 +11,9 @@ module.exports = {
           type: Sequelize.INTEGER
         },
 
-        <% _.each(attributes, function (dataType, fieldName) { %>
-          <%= fieldName %>: {
-            type: Sequelize.<%= dataType.toUpperCase() %>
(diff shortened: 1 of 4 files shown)


link-alex:
Ok, so I forgot that String.prototype.includes appeared in node v6, so will update to support node v4+

link-alex:
@sushantdhiman @sdepold could please take a look?

src/assets/migrations/create-table.js
```diff
@@ -11,9 +11,9 @@ module.exports = {
           type: Sequelize.INTEGER
         },
 
-        <% _.each(attributes, function (dataType, fieldName) { %>
-          <%= fieldName %>: {
-            type: Sequelize.<%= dataType.toUpperCase() %>
+        <% _.each(attributes, function (attribute) { %>
```
sushantdhiman: `attributes` is an array, right? We can use `.forEach` here to simplify

link-alex:
@sushantdhiman updated :)

grim2888:
thank you

Issue #590, opened 2017-11-23
Bug? db:create cannot set default charset with connection information.


ALiangLiang:
## What you are doing?

```js
// config/config.js
require('dotenv').config()
module.exports = {
  development: {
    username:: process.env.DEV_DB_USER,
    password: process.env.DEV_DB_PASS,
    database: process.env.DEV_DB_NAME, //database_dev
    host: process.env.DEV_DB_HOST,
    dialect: process.env.DEV_DB_TYPE,
    dialectOptions: {
      charset: 'utf8mb4',
      collate: 'utf8mb4_general_ci'
    },
    logging: true
  }
}
```

```bash
$ sequelize db:create --env development
```

## What do you expect to happen?
Create a database with charset in config/config.js
```bash
Executing (default): CREATE DATABASE `database_dev` DEFAULT CHARSET=`utf8mb4 DEFAULT COLLATE=`utf8mb4_general_ci`
```

## What is actually happening?
Doesn't use this configure.
```bash
# log
Executing (default): CREATE DATABASE `database_dev`
```

__Dialect:__ mysql
__Database version:__ 5.7.20
__Sequelize CLI version:__ 3.1.0
__Sequelize version:__ 4.22.15

chris-fung:
Met the same problem with yours. Just need to modify the charset and collate by myself after creating db

galki:
according to [this](https://stackoverflow.com/questions/41666820/node-sequelize-mysql-how-to-define-collation-and-charset-to-models) using `define` instead of `dialectOptions` works when creating a sequelize connection, but it doesn't work when using cli to `db:create`

i also tried to set the charset/collation in the model migration [like the docs say](http://docs.sequelizejs.com/class/lib/query-interface.js~QueryInterface.html#instance-method-createTable) by adding the third arg to:
```
queryInterface.createTable(
  tableName,
  { /*... columns*/ },
  { charset: 'utf8' },
)
```
and that results in error:
```
ERROR: Can't create table `mydb`.`mytable` (errno: 150 "Foreign key constraint is incorrectly formed")
```

kwbtdisk:
I avoid this problem by querying `ALTER DATABASE` from a migration file like this:

`/db/migrations/00000-fix-db-charset.js`
```javascript
module.exports = {
  up: (queryInterface, Sequelize) => {
    return Promise.all([
      queryInterface.sequelize.query(
        `ALTER DATABASE ${queryInterface.sequelize.config.database}
[...]

PR #591, opened 2017-11-23
fix(db:create/drop): Add charset parameter in query string (#590)

ALiangLiang:
for #590

### DIFF
--- a/src/commands/database.js
+++ b/src/commands/database.js
@@ -19,7 +19,10 @@ exports.handler = async function (args) {
 
   switch (command) {
     case 'db:create':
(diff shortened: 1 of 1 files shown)


sushantdhiman:
Sorry I can accept this. This is just too much complexity for CLI. I just added `db:create/db:drop` because of huge demand.

If you want this to be added, please open a PR to sequelize and add `createDatabase` / `dropDatabase` to query-interface which can properly parse additional options.

Then whenever those methods are available CLI can just call them

ALiangLiang:
OK.  I see. I agree add `createDatabase` / `dropDatabase` is better solution.

PR #620, opened 2018-02-14
Model association for sequalize-cli 4.0.0 updates

buddylindsey:
Created this to deal with a merge conflict so it can be merged. Preserved commits for contribution credit.

### DIFF
--- a/test/model/create.test.js
+++ b/test/model/create.test.js
@@ -210,6 +210,7 @@ const _         = require('lodash');
                     .src(Support.resolveSupportPath('tmp', 'models'))
                     .pipe(helpers.readFile('user.js'))
                     .pipe(helpers.ensureContent(targetContent))
+                    .pipe(helpers.ensureContent('.associate'))
                     .pipe(helpers.teardown(done));
                 });
               });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 All notable changes to this project will be documented in this file.
 
 ## Future
+## v3.2.0 - 3rd, Dec 2017
+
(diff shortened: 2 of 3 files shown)


buddylindsey:
@sushantdhiman I think the tests will pass if they can just be restarted. Looks like one got hung up for a very long time. Are you able to restart the build? I don't see a button to do it so I don't think I have permissions. If not I can try to find something to push to kick off another build.

sushantdhiman:
No problem, I will retry that test

test/model/create.test.js
```diff
@@ -197,7 +197,7 @@ const _         = require('lodash');
 
                 const targetContent = attrUnd.underscored ?
                   'underscored: true'
-                  : '{\n    classMethods';
+                  : '{});';
```
sushantdhiman: assert `associate`

test/model/create.test.js
```diff
@@ -210,7 +210,8 @@ const _         = require('lodash');
                     .src(Support.resolveSupportPath('tmp', 'models'))
                     .pipe(helpers.readFile('user.js'))
                     .pipe(helpers.ensureContent(targetContent))
-                    .pipe(helpers.teardown(done));
+                    .pipe(helpers.teardown(done))
+                    .pipe(helpers.ensureContent('.associate'));
```
sushantdhiman: `.pipe(helpers.ensureContent('.associate'))`, Should come before `teardown`, I think tests will fail

buddylindsey:
@sushantdhiman looks like we are good to go. It is now mergable and the tests have all passed.

Issue #284, opened 2016-04-07
cli appending "public" schema in front of database name during migrations.


buildjosh:
So I'm using the latest cli and latest sequelize.

Config file looks like this:

```json
{
  "development": {
    "use_env_variable": "DATABASE_URL",
    "dialect": "postgres"
  },
  "production": {
    "use_env_variable": "DATABASE_URL",
    "dialect": "postgres"
  }
}
```

DATABASE_URL looks like:

```
postgres://josh@localhost:5432/crm
```

When trying to run migrations:

```sh
$  ./node_modules/.bin/sequelize db:migrate
```

I'm getting this:

```
{ [SequelizeDatabaseError: relation "public.crm.public.accounts" does not exist]
  name: 'SequelizeDatabaseError',
  message: 'relation "public.crm.public.accounts" does not exist',
  parent: 
   { [error: relation "public.crm.public.accounts" does not exist]
     name: 'error',
     length: 117,
     severity: 'ERROR',
     code: '42P01',
     detail: undefined,
     hint: undefined,
     position: undefined,
     internalPosition: undefined,
     internalQuery: undefined,
     where: undefined,
     schema: undefined,
     table: undefined,
     column: undefined,
     dataType: undefined,
     constraint: undefined,
     file: 'namespace.c',
     line: '413',
     routine: 'RangeVarGetRelidExtended',
     sql: 'ALTER TABLE "public"."crm.public.accounts" ADD COLUMN "agentId" VARCHAR(255) NOT NULL DEFAULT \'NOT_TRACKED\';' },
  original: 
   { [error: relation "public.crm.public.accounts" does not exist]
     name: 'error',
     length: 117,
     severity: 'ERROR',
     code: '42P01',
     detail: undefined,
     hint: undefined,
     position: undefined,
     internalPosition: undefined,
     internalQuery: undefined,
     where: undefined,
     schema: undefined,
     table: undefined,
     column: undefined,
     dataType: undefined,
     constraint: undefined,
     file: 'namespace.c',
     line: '413',
     routine: 'RangeVarGetRelidExtended',
     sql: 'ALTER TABLE "public"."crm.public.accounts" ADD COLUMN "agentId" VARCHAR(255) NOT NULL DEFAULT \'NOT_TRACKED\';' },
  sql: 'ALTER TABLE "public"."crm.public.accounts" ADD COLUMN "agentId" VARCHAR(255) NOT NULL DEFAULT \'NOT_TRACKED\';' }

```

[...]

PR #622, opened 2018-02-16
Add option for searchPath when the connection is set by URL

hisorange:
Minor addition which is useful when migrating into multiple schemas.
We had an issue when the same migration is needed in dozens of schemas, already applied and using this simple solution. 

It's not breaking anything but adds a way to solve issue like #284 and any other multi tenant application ^.^

### DIFF
--- a/src/helpers/config-helper.js
+++ b/src/helpers/config-helper.js
@@ -158,19 +158,29 @@ const api = {
 
   urlStringToConfigHash (urlString) {
     try {
-      const urlParts = url.parse(urlString);
+      const urlParts = url.parse(urlString, true);
       let result   = {
         database: urlParts.pathname.replace(/^\//,  ''),
         host:     urlParts.hostname,
         port:     urlParts.port,
         protocol: urlParts.protocol.replace(/:$/, ''),
-        ssl:      urlParts.query ? urlParts.query.indexOf('ssl=true') >= 0 : false
+        ssl:      (urlParts.query && urlParts.query.ssl === 'true')
       };
 
       if (urlParts.auth) {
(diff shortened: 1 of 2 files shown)


sushantdhiman:
Need to add a test here https://github.com/sequelize/cli/blob/master/test/url.test.js

src/helpers/config-helper.js
```diff
@@ -174,6 +174,16 @@ const api = {
         });
       }
 
+      // If there is a search path key in the URL, then the prependSearchPath is being activated.
+      if (urlParts.protocol === 'postgres' && urlParts.searchPath) {
+        result = _.assign(result, {
+          searchPath:     urlParts.searchPath,
```
sushantdhiman: Get rid of extra space between `searchPath:` and `urlParts.searchPath`

src/helpers/config-helper.js
```diff
@@ -158,13 +158,13 @@ const api = {
...
     try {
-      const urlParts = url.parse(urlString);
+      const urlParts = url.parse(urlString, true);
       let result   = {
         database: urlParts.pathname.replace(/^\//,  ''),
         host:     urlParts.hostname,
         port:     urlParts.port,
         protocol: urlParts.protocol.replace(/:$/, ''),
-        ssl:      urlParts.query ? urlParts.query.indexOf('ssl=true') >= 0 : false
+        ssl:      (urlParts.query && urlParts.query.ssl === 'true')
```
sushantdhiman: `urlParts.query && urlParts.query.ssl === 'true' || false`

Will need to add test for this change as well

Issue #265, opened 2016-02-25
Postgres schema for migration table


angelstoone:
Hi guys,

I would like to create the migration table in a given postgres **postgres schema** different from the **public** one. Is there an option in the `config.json` file to specify the **postgres schema** of the migrations table?

I am talking about something like the `migrationStorageTableName` option, for example, called `migrationStorageSchemaName`?

Cheers

Americas:
Hi @angelstoone, currently there isn't such an option, but it is doable.

A pull request is greatly appreciated if you are up to the task. If not, I can get to this in the not too distant future :laughing:

angelstoone:
Hi @Americas, it looks like someone else already did it #141 but it has not been merged yet.

Americas:
PR #205 seems to do the trick. However, it needs to be updated, and requires tests!

jspeis:
+1 for supporting this feature

llonchj:
+1

TizianoPerrucci:
+1

webloem:
+1

sushantdhiman:
Fixed by https://github.com/sequelize/cli/pull/635

PR #635, opened 2018-03-26
Allow the Postgres schema for the migration table to be specified through migrationStorageTableSchema (#265)

kpeng:
Thanks to @jonnolen's original PR for getting me started on this. Added a test that injects `migrationStorageTableSchema` to a test config and ensures that the non-public schema is created correctly.


kpeng:
Unsure what's causing the schema to not be created here, both on my local PostgreSQL instance and a Docker instance, all of the tests pass.

kpeng:
Hi, pinging to try to get some more information. Still unsure why it's failing on TravisCI, since I have tests passing locally off a fresh Postgres instance bootstrapped by the Docker Compose configuration. Any ideas?

src/core/migrator.js
```diff
@@ -61,6 +61,24 @@ export function getMigrator (type, args) {
 
     return sequelize
       .authenticate()
+      .then(() => {
+        // Check if this is a PostgreSQL run and if there is a custom schema specified, and if there is, check if it's
+        // been created. If not, attempt to create it.
+        if (helpers.version.getDialectName() === 'pg') {
+          const customSchemaName = helpers.umzug.getSchema('migration');
+          if (customSchemaName && customSchemaName !== 'public') {
+            return sequelize.showAllSchemas().then(schemas => {
```
sushantdhiman: Use `createSchema` directly which handles existing schema with `IF NOT EXISTS`  https://github.com/sequelize/sequelize/blob/aa92764f953eeefbf194a6e2b765552adfa73bda/lib/query-interface.js#L35

test/db/migrate.test.js
```diff
@@ -143,6 +143,35 @@ const _         = require('lodash');
         });
       });
     });
+
+    describe('custom meta schema', () => {
+      it('correctly uses the defined schema', function (done) {
+        const self = this;
```
sushantdhiman: `self = this` not required with arrow functions

test/support/index.js
```diff
@@ -90,27 +91,55 @@ var Support = {
...
+              });
+        })
+        .catch(function (err) {
+          console.log('Error in support.clearDatabase() dropAllTables() :: ', err);
+        });
+    };
+
+    // If Postgres, loop through each of the non-public schemas and DROP/re-CREATE them.
+    if (this.dialectIsPostgres()) {
+      sequelize
```
sushantdhiman: I think tests are failing because of missing return

test/support/index.js
```diff
@@ -90,27 +91,55 @@ var Support = {
...
+        .showAllSchemas()
+        .then(function (schemas) {
+          // showAllSchemas() leaves off the public schema.
+          let searchPathSchemas = schemas.join(', ');
+          let promise = Bluebird.resolve();
+
+          _.forEach(schemas, function (schema) {
```
sushantdhiman: Use `Promise.map` + `Promise.all` combination rather than `forEach` for better readability

test/support/index.js
```diff
@@ -90,27 +91,55 @@ var Support = {
...
+                });
             });
-      })
-      .catch(function (err) {
-        console.log('Error in support.clearDatabase() dropAllTables() :: ', err);
-      });
+          });
+
+          // Drop the public schema tables.
+          promise.then(dropAllTables);
```
sushantdhiman: return the promise

kpeng:
@sushantdhiman, thanks for your comments, updated the PR with fixes. Ran the tests locally and they pass under the same node and npm versions, the main difference is that I'm running PostgreSQL 10.1. I can't think of why that would cause the failure. The other possibility is a PG permission/role issue with the created schema?

sushantdhiman:
>The other possibility is a PG permission/role issue with the created schema?

@kpeng Yeah I ran your PR locally and tests pass properly with Postgres 9.5 as well. Looks like permission issue with schema creation

sushantdhiman:
@kpeng Please rebase this branch with `master`, We are now using docker setup for CI as well, let's see if that works

kpeng:
Thanks, @sushantdhiman, that seemed to have worked!

sushantdhiman:
Thanks @kpeng , Released as `v5.1.0`

Issue #651, opened 2018-05-22
Support import/export Statements Using ESM


rtablada:
There has been a lot of discussion around Babel transforms in #189 for years and there doesn't seem to be much progress since babel is fairly environment/team specific in its implementation.

From reading through the thread, the major theme seems to be more around `import` and `export` syntax.
This can be solved using ESM which is a production ready and well supported module loader for Node.

An `--esm` flag or `esm` true option could be added to CLI to support ESM loading for CLI commands.

A more flexible option could be to allow users to pass in their own `require` definition that could be used in the CLI for project requires and module loading although this could cause some 🐉s to pop up.

rtablada:
I hope to work on a proof of concept for this soon, but wanted to put out an issue to gauge interest as well.

sonufrienko:
I use ESM to load ES Modules instead of Babel.
So this is a wanted feature.

Thanks, @rtablada  :+1:

SpiritualDixit:
Any update on it?

mariusa:
`sequelize-cli migration:generate` should also generate migrations with ESM syntax when project package.json already has `type=module`

Issue #112, opened 2015-03-27
Work with babel


fundon:
Hi, currently I write ES6+ code and use [Babel](http://babeljs.io/) transform them.

Maybe we can include it and add an option for it.

Like this below:

``` js

# lib/gulpfile.js

"use strict";

require("babel/register");

var gulp    = require("gulp");
var path    = require("path");
var fs      = require("fs");
var helpers = require(path.resolve(__dirname, "helpers"));

require("gulp-help")(gulp, {
  aliases: ["h"],
  afterPrintCallback: helpers.gulp.printManuals
});

...

```

If you like this, I can submit a PR.

sdepold:
While I don't really need ES6 for the CLI at this point, I wonder if the end-user will benefit from such a change?

fundon:
I refactor my project and use ES6+ by Babel.
The database.js below and the migrations/*.js files will do that.
So I need babel to transform code before task execute.


```js
export default {

  "development": {
    "url": process.env.DATABASE_URL,
    "username": "trek",
    "password": "star trek",
    "database": "trek",
    "host": "192.168.59.103",
    "port": "5432",
    "dialect": "postgres",
    "native": true,
    "timezone": "+00:00",

    "query": {
      "pool": true,
      "debug": true
    },

    "define": {
      "timestamps": true,
      "underscored": true,
      "charset": "utf8"
    },
    "logging": true
  },

  "test": {
    "url": process.env.DATABASE_URL,
    "username": "root",
    "password": null,
    "database": "database_test",
    "host": "127.0.0.1",
    "dialect": "postgres"
  },

  "production": {
    "url": process.env.DATABASE_URL,
    "username": process.env.DATABASE_USERNAME,
    "password": process.env.DATABASE_PASSWORD,
    "database": process.env.DATABASE_DATABASE,
    "host": process.env.DATABASE_HOST,
    "port": process.env.DATABASE_PORT,
    "dialect": "postgres"
  }

};
```

fundon:
BTW, likes the `--coffee` flag.

xpepermint:
+1

xpepermint:
@sdepold Any progress on this? My projects are all es6 (babel). What do you suggest?

xpepermint:
Btw... I think it would be enough if you just add `require("babel/register");` into `lib/gulpfile.js` and `bin/sequelize`. I made a fork https://github.com/xpepermint/cli just to test that and it works but I think you should check if that's true.

[...]

PR #654, opened 2018-05-23
Support esm for module loading using option flag

rtablada:
This PR adds support to use [ESM](https://github.com/standard-things/esm) to load ES6 modules as mentioned in #651 and would possibly be a resolution for a lot of people in #112 who's main concern is loading modules and not full transpilation.

I tested on my codebase and this works for `db:create`, `db:drop`.

~However further updates to `umzug` will be needed to support this for to work for seeds and migrations.~


rtablada:
Because of ESM support engines this requires node version >6.

Because of node support as mentioned in #655 this should be the support stack going forward.

rtablada:
After looking through `umzug` source, by passing a `customResolver` function using ESM, then ESM modules are supported and work for all migration and seed commands.

src/commands/migrate.js
```diff
@@ -83,7 +83,7 @@ function migrationStatus(args) {
   }).catch(e => helpers.view.error(e));
 }
 
-function migrateSchemaTimestampAdd(args) {
+function migrateSchemaTimestampAdd (args) {
```
ebryn: @rtablada you want to remove these unnecessary whitespace changes?

rtablada: The changes seem to be from the eslint config in the project that was missing. I can try to turn off autoformat/fix in my config.

sushantdhiman:
I am reluctant to add this change as it doesn't actually introduce anything of significant use apart from ES6/7 style `import`.

Please check https://github.com/sequelize/cli/issues/112#issuecomment-394172092 / [docs](https://github.com/sequelize/sequelize/blob/master/docs/migrations.md#using-babel) to see how you can use `babel` with `sequelize-cli` setup, which will allow you to use full ES6/7 setup without much friction.

I may include this change in future CLI v5, but chances are slim.

Thanks for contributing to CLI. If my solution doesn't work for you, please let me know.

ebryn:
> it doesn't actually introduce anything of significant use apart from ES6/7 style import.

@sushantdhiman that's the whole point. you don't consider ES modules significantly useful?

ebryn:
@sushantdhiman i believe it's also fair to say that you should be recommending using `@std/esm` over `babel` nowadays. maybe @jdalton can chime in?

jdalton:
Hi @ebryn!

Instead of baking in `esm` support you could add a `--require` (`-r` alias) this would allow `esm`, `@babel/register`, `ts/register`, and others to be used.

sonufrienko:
I use ESM to load ES Modules instead of Babel.
So this is a wanted feature.

Thanks, @rtablada  :+1:

Issue #658, opened 2018-06-04
Make timestamp optional


adeelhussain:
## What you are doing?
```js
sequelize model:generate --attributes as --name modelName --attributes firstName:string
```

## What do you expect to happen?
I would like to make the timestamp optional because for some tables, there is no need of that.

## What is actually happening?
There should be a flag something like
```js
sequelize model:generate --name modelName --attributes firstName:string --timestamp false
```
which will skip the timestamp

adeelhussain:
@papb can you please have a look on this issue, PR of this is here https://github.com/sequelize/cli/pull/665

papb:
Hello, I am super time constrained right now, can you please ping me again in another week or so?

PR #665, opened 2018-06-21
Feature: Make timestamp optional #658

adeelhussain:
Making timestamp optional, simply pass `--timestamp false` will skip the timestamp from the migration file and set the `timestamps: false` in model
Referenced to #658


src/commands/model_generate.js
```diff
@@ -17,6 +17,12 @@ exports.builder =
           type: 'string',
           demandOption: true
         })
+        .option('timestamp', {
```
sushantdhiman: `timestamps`

src/assets/migrations/create-table.js
```diff
@@ -11,21 +11,30 @@ module.exports = {
...
         },
 
-        <% attributes.forEach(function(attribute) { %>
+        <% attributes.forEach(function(attribute, index) { %>
           <%= attribute.fieldName %>: {
             type: Sequelize.<%= attribute.dataFunction ? `${attribute.dataFunction.toUpperCase()}(Sequelize.${attribute.dataType.toUpperCase()})` : attribute.dataValues ? `${attribute.dataType.toUpperCase()}(${attribute.dataValues})` : attribute.dataType.toUpperCase() %>
-          },
+          }
+          
+          <% if(attributes.length -1 !== index) { %>
```
sushantdhiman: space after `if`

src/assets/migrations/create-table.js
```diff
@@ -11,21 +11,30 @@ module.exports = {
...
+            ,
+          <% } %>
+
         <% }) %>
 
-        <%= createdAt %>: {
-          allowNull: false,
-          type: Sequelize.DATE
-        },
+        <% if(typeof(createdAt) != "undefined"){ %>
```
sushantdhiman: `if (typeof createdAt !== undefined)`

src/commands/model_generate.js
adeelhussain: @sushantdhiman done!

adeelhussain:
Well i think its not a problem but i have removed that trailing comma.

sushantdhiman:
Build/Tests are still failing

ephys:
In case OP becomes active with this PR again:

Options should be false by default. If we want timestamps to generate by default, the cli option to disable them should be named `skip-timestamps` or `no-timestamps` and default to false.

I'm also closing the related documentation PR as this one is closed. (https://github.com/sequelize/sequelize/pull/11410)
If activity resumes, we can reopen both PRs

Issue #663, opened 2018-06-20
Filename printed for model:generate doesn't match path on disk


aguynamedben:
## What you are doing?

Generating a model using `model:generate`

```
$ node_modules/.bin/sequelize model:generate --name Doc --attributes id:uuid:primary,upstream_id:string,domain:string,source:string,title:string,url:string
```

![image](https://user-images.githubusercontent.com/107841/41631578-3b348748-73ea-11e8-9589-0cfa738cfe75.png)

## What do you expect to happen?
The filename in the output text to end with `-create-doc.js` (what's actually written to disk)

## What is actually happening?
The filename in the output text ended with `-Doc.js` (that file doesn't exist!)

__Dialect:__ sqlite
__Database version:__ 3.24.0 2018-06-04 19:24:41 c7ee0833225bfd8c5ec2f9bf62b97c4e04d03bd9566366d5221ac8fb199a87ca
__Sequelize CLI version:__ sequelize-cli@4.0.0
__Sequelize version:__ sequelize@4.37.10

aguynamedben:
BTW, thanks for this great library, I'm new to it and finding it very useful 🙏

adeelhussain:
@sushantdhiman I have created a [PR](https://github.com/sequelize/cli/pull/669) for this issue, can you please have a look?
#669

sushantdhiman:
Fixed by https://github.com/sequelize/cli/commit/c4e53b18b4426f3614b91a949daeb28c4d95f913

PR #669, opened 2018-06-27
Bug Fix: model:generate doesn't match path on disk #663

adeelhussain:
Fixed bug of printing wrong path of generated migration #663

### DIFF
--- a/src/helpers/path-helper.js
+++ b/src/helpers/path-helper.js
@@ -55,7 +55,7 @@ module.exports = {
   },
 
   getMigrationPath (migrationName) {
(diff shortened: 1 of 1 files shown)


aguynamedben:
I didn't run the code myself, but it seems like this PR only addresses the capitalization issue? The original issue also identified a naming issue as well (`-create-doc.js` vs. `-Doc.js` suffix).

The code itself looks looks good to me, just making sure you caught that aspect of the issue. 👍

sushantdhiman:
Fixed by https://github.com/sequelize/cli/commit/c4e53b18b4426f3614b91a949daeb28c4d95f913

PR #680, opened 2018-08-04
Adds in force flag for model:create

### LINKED ISSUE #638: Command argument --force does not work with model:generate
## What you are doing?

I am running this command
```
sequelize model:generate --name experiment --atributes col1:text,col2:text
```
However, as I already have an experiment.js file in models I get the error.
`The file already exists. Run command with --force to overwrite it.`

So I run
`sequelize model:generate --force --name experiment --atributes col1:text,col2:text`

I expected the model to update.

## What is actually happening?
I get "unknown argument: force".

__Dialect:__ postgres
__Database version:__ 9.6
__Sequelize CLI version:__ 4.0.0
__Sequelize version:__ 4.37.6

Projjol:
Currently if model:create tries to recreate the migration and model
files for the same name, the user is asked to use the --force flag which
does not currently work as it is not being accepted currently. This
allows --force with value true

Fixes #638

### DIFF
--- a/src/commands/model_generate.js
+++ b/src/commands/model_generate.js
@@ -17,6 +17,11 @@ exports.builder =
           type: 'string',
           demandOption: true
         })
+        .option('force', {
+          describe: 'Forcefully re-creates model with the same name',
+          type: 'string',
+          demandOption: false
+        })
     )
       .help()
       .argv;


sushantdhiman:
Thanks

PR #682, opened 2018-08-14
Add model:generate attribute datatype validation

Georift:
I ran into a small problem today when generating some models, it's pretty easy to mistakenly write something like 'someDate:datetime' as an attribute when it is just 'someDate:date'. The generate command lets it slide and then you end up with some [hard to debug .toString() errors](https://stackoverflow.com/questions/46034605/sequelize-cannot-read-property-tostring-of-undefined) by when migrating.

I've added an error condition which halts the generation if any of the attributes contain an invalid DataType.

Also, I noticed that `runCli()` in `test/support/helpers.js` wasn't correctly asserting a non-zero error code so I fixed it by not passing through result.

### DIFF
--- a/src/commands/model_generate.js
+++ b/src/commands/model_generate.js
@@ -31,7 +31,13 @@ exports.handler = function (args) {
   ensureMigrationsFolder();
   checkModelFileExistence(args);
 
-  helpers.model.generateFile(args);
+
+  try {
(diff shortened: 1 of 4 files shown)


sushantdhiman:
Thanks this fixed https://github.com/sequelize/cli/issues/222

Issue #693, opened 2018-09-18
Feature request: Natively run Typescript migrations and seeders


mgibson91:
In the same way that current CLI works, it would be great to be able to consume Typescript files.

My current project heavily uses aliases paths in development, runtime and testing using:
- Dev: tsconfig paths
- Test: Jest moduleNameMapper
- Production: tsnode and tsconfigpaths

It's extremely pleasant.The amount of different solutions there are (which can also be a nightmare to get configured in all three execution contexts) give an indication to the importance and popularity of aliased paths.

Currently, to reuse existing code, I write my migrations, seeders and models all in Typescript, then transpile the migrations and seeders to JS so the sequelize cli can work its magic.

However, the pain comes when the transpiles JS has not converted the aliased paths to relative paths and understandably, the Sequelize migrator cannot find these aliased paths.

It would be so good to be able to natively run the Typescript migrations and seeders, perhaps via ts-node and tsconfig-paths so that a user can reuse code in the same way the do in the rest of their project with no extra work.

I expect it would require the migration code to be converted to Typescript and then be executed as:

ts-node -r tsconfig-paths/register migrator.ts
OR
node --require ts-node/register -r tsconfig-paths/register migrator.ts

raphaelsoul:
Since v5.0+ supports TS natively. I do really think it is a must. 
I am writing model in TS but migrations and seeders in es6. The most problem is I cannot use enum type in seeder and migrations, which is used to define the possible values for some column.

And is there any workaround for now?

TranBaVinhSon:
It seems like we have to deal manually with this problem by ourself. 🤔

ghosh-jaideep:
@sdepold @sushantdhiman @Americas 
We are using this feature along with feathersjs (ts), But the sequelize-clli generates the js files only, So help us to configure it to generate ts instead of js migration files.

jaideepghosh:
@sdepold @sushantdhiman @Americas 
We are using this feature along with feathersjs (ts), But the sequelize-clli generates the js files only, So help us to configure it to generate ts instead of js migration files.

kunwar97:
You can always change the path of migrations & seeders to point dist directory

`module.exports = {
  'config':          path.resolve('dist', 'config', 'database.js'),
  'migrations-path': path.resolve('dist', 'migrations'),
  'models-path':     path.resolve('src', 'models'),
  'seeders-path':    path.resolve('dist', 'seeders'),
}`

iMuFeng:
[...]

PR #715, opened 2018-11-09
Allow both ts and js files to run migrations

orta:
Because you can use `ts-node node_modules/.bin/sequelize db:migrate` to trigger your migrations, you can safely use TypeScript using more or less vanilla sequelize-cli. The only blocker I've found so far is the check for js files in the migration file detection. 

This is great, because you get awesome tooling support

screen shot 2018-11-08 at 7 14 16 pm

This PR allows a `.ts` file as well as a `.js` to pass by default.

```
> "hello.ts".match(/\.(t|j)s$/)
< [".ts", "t"] (2)
> "hello.js".match(/\.(t|j)s$/)
< [".js", "j"] (2)
> "hello.coffee".match(/\.(t|j)s$/)
< null
```

There's a downside to this in that someone could try running a typescript file without the `ts-node`, but this app is not documented as supporting typescript out of the box, so I don't think people would have that expectation.

### DIFF
--- a/src/core/migrator.js
+++ b/src/core/migrator.js
@@ -48,7 +48,7 @@ export function getMigrator (type, args) {
       migrations: {
         params: [sequelize.getQueryInterface(), Sequelize],
         path: helpers.path.getPath(type),
-        pattern: /\.js$/,
+        pattern: /\.(t|j)s$/,
         wrap: fun => {
           if (fun.length === 3) {
             return Bluebird.promisify(fun);


mkay581:
Just ran into the `File does not match pattern: /\.js$/` error when attempting to run the seed and migrate commands using `ts-node-dev`. This should fix that, so thank you!!!

Will this change also work for the `seed` command and when running a ts file as the config?

```bash
sequelize db:seed:all --config=./config/sequelize-config.ts
```

mkay581:
Also look likes this would resolve #693, if merged. Hopefully this would be accepted soon.

mkay581:
@orta did you create a fork of this that can be used until this PR is merged? Really could use this. This is a very tiny change that is fully backwards-compatible so not sure why it's been sitting for so long.

xianxingg:
This is a temp solution.
If you want to use typescript, use it from the beginning to the end. That's to say, the option `--typescript` should be supported for generating migration files, and cli should do the job of creating ".ts" file with "export default = { ... }" as content.

lubomirigonda:
Hi I'd like to ask when the PR will be accept ? It is quite a while :/. I tried this fix and it look that is working very well

TranBaVinhSon:
Hmm, why this PR still not merged?

papb:
Hi, us maintainers have been focusing on the main repo recently, but I intend to take a good look on the CLI in the near future. Thanks everyone for the patience!

brianemoore57:
One possiblity is to allow the user to add his own check for js files or globbing expression. A related issue would be the ability to suppress the 'does not match pattern: /\.js$/   We have a large project and wish to use subdirectories like /migrations/data and seed/data but it results in a lot of distracting messages.

max10rogerio:
I forked this repository and tried this solution! And works!

I just added the script on packege:
`node -r ts-node/register ./node_modules/sequelize-cli/lib/sequelize`

> example
```ts
import { QueryInterface, DataTypes, literal } from 'sequelize'

export function up(queryInterface: QueryInterface): Promise<void> {
  return queryInterface.createTable('Estados', {
    id: {
      allowNull: false,
      autoIncrement: true,
      type: DataTypes.INTEGER,
      comment: 'Identificador da tabela Estados',
      primaryKey: true,
    },
    nome: {
      allowNull: false,
      type: DataTypes.STRING(20),
      comment: 'Nome do estado',
    },
    sigla: {
      allowNull: false,
      type: DataTypes.STRING(2),
      comment: 'Sigla do Estado',
    },
    createdAt: {
      allowNull: false,
      type: DataTypes.DATE,
      comment: 'Data da criação do registro',
      defaultValue: literal('NOW()'),
    },
    updatedAt: {
      allowNull: false,
      type: DataTypes.DATE,
      comment: 'Data da última alteração do registro',
      defaultValue: literal('NOW()'),
    },
    deletedAt: {
      allowNull: true,
      type: DataTypes.DATE,
      comment: 'Data da exclusão do registro',
    },
  })
}

export function down(queryInterface: QueryInterface): Promise<void> {
  return queryInterface.dropTable('Estados')
}

```

cjancsar:
@papb can this be merged?

ismail-codinglab:
All checks passed, no conflicts

why not merge?

ismail-codinglab:
Looking at the main contributors

@sdepold @sushantdhiman @Americas

kf6kjg:
The problem I would have with this commit is that it allows the following files to be considered different migrations:
```
fixstage.ts
fixstage.js
```
To me those are the same migration, just one was compiled from the other.  Compiling the TS into the same folder is indeed very bad form, but it still is a logical problem.

Also I'd personally, though I'm by no means a maintainer of this project, that a PR exist for adding a flag that allows the user to specify the matching pattern or file extension with the default left as it stands in the project currently.  That way the user can override but the default is left unaffected.

Lastly, consider the principle of least surprise: which is worse, finding that after you upgraded your tools that all of a sudden both the TS and JS migrations files are being executed and you are being hit with failures left and right, or that an option now exists to change the pattern so that your TS files are run? I'd argue the latter is the least surprising and thus better.

namnm:
I have been able to work around this using a postinstall script to run this snippet:
```js
const { readFileSync, writeFileSync } = require('fs')
const { join } = require('path')

// Fix sequelize-cli doesnt support typescript extensions
// Will be run in the postinstall script in package.json
const p = join(__dirname, '../node_modules/sequelize-cli/lib/core/migrator.js')
const f = readFileSync(p, 'utf8').replace('.js$/', '.(j|t)sx?$/')
writeFileSync(p, f, 'utf8')
```

mohsen1:
@namnm `patch-package` npm package is a more clean hacky solution

psi-4ward:
push bump +1 :)

psi-4ward:
So the best option would be a config setting in `.sequelizerc` !

kf6kjg:
I switched my project over to Umzug, which is the sequelize library that this project uses to do migrations. Gave me a lot more fine grained control over the process and allowed for my application to run migrations on startup. 
A recent addition to Umzug solves the problem that this PR attempts by simply making the file extension not part of the comparison and not stored.

MBerka:
TS migration blues led me to the same decision earlier today. In case anyone else comes looking for a way to get a ts-node Mocha integration test to load their TS migrations despite those migrations having no JS counterparts during testing because ts-node does not output files and it's all buried in a container - the simplest approach with umzug uses an arbitrary pattern parameter, `{ pattern: /^\d+[\w-]+\.(js|ts)$/ }`.
That's a more rigorous pattern too; there could be multiple use cases to keep  the default "js" pattern, and expose an optional pattern input, not just to admit TS, but to block arbitrary other files. This being CLI, maybe as a string passed to RegExp()? It's a heavier change though.

mkay581:
> I switched my project over to Umzug

@kf6kjg This is an option but AFAIK I don't think umzug does seeding out of the box like this library? You could do some manual work to have umzug do seeding as evidence by [this comment](https://github.com/sequelize/umzug/issues/24#issuecomment-591313032), but its a brittle workaround with a few caveats.

kf6kjg:
@mkay581 in Umzug seeders are just another set of migrations. You are totally in control of how you want to structure all parts of your migration: you could have schema migrations, seeds, data migrations, and any other migration type your project needs just by running Umzug against either different folders, different file name patterns, or both. 

Of course Umzug isn’t a cli interface, but I really didn’t need a cli interface in my projects.

mkay581:
Sure, although some people would rather not have to learn such granular level of detail, which is why people choose to use this higher level library.

ismail-codinglab:
If you would ask me what would I do different if I would start a typescript project again is that I would use typeorm. Sequelize isn't build for typescript, it's Javascript conversion to typescript

mkay581:
> it's Javascript conversion to typescript

JavaScript doesnt get converted to TypeScript

MBerka:
@mkay581 Not automatically, no, only when you start with a plain JS project and later manually migrate the project to TS, migrations and all, learning all the typing tricks needed to make it keep working along the way. I think that's the point - that TypeORM is TS-native.
That said, I was interested to find [sequelize-typescript](https://www.npmjs.com/package/sequelize-typescript) that makes Sequelize's models look rather similar.

kroleg:
@sushantdhiman can we please include this in v6? It's a super easy and long awaited change. And `cjs` support was just added.

kroleg:
@sushantdhiman can you please explain why did you close this one?

orta:
Screen Shot 2020-07-03 at 10 55 15 AM

All PRs were closed ^

sushantdhiman:
PR Cleanup, if anyone needs this feature they should open a new PR & add a test for this. For example https://github.com/sequelize/cli/pull/905

yonycalsin:
```bash
npx ts-node -T -r tsconfig-paths/register .\node_modules\sequelize-cli\lib\sequelize db:migrate
```

Issue #172, opened 2015-08-25
Logging sensitive info when using connection url.


juanjoLenero:
I'm using the database connection url as an alternative to the config file. Whenever I run a migration, gulp outputs some logs that expose the url. Is there a way to avoid this without the config file?

Example logs:
```
> com.***@1.0.0 migrate /Users/juanjo/development/web/backend
> sequelize --url postgres://$DB_USERNAME:$DB_PASSWORD@$DB_HOST/$DB_NAME db:migrate


Sequelize [Node: 0.12.0, CLI: 1.7.4, ORM: 3.6.0, pg: ^4.2.0]

Parsed url postgres://user:password@localhost/database *** <--- Hide this?
Using gulpfile ~/development/web/backend/node_modules/sequelize-cli/lib/gulpfile.js
Starting 'db:migrate'...
Finished 'db:migrate' after 75 ms
No migrations were executed, database schema was already up to date.
```

sdepold:
It'S currently not build in but should be an easy thing to remove. Would you like to provide a PR?

sdepold:
@juanjoLenero is that change good enough? https://github.com/sequelize/cli/pull/164

juanjoLenero:
Yes, that'll work, thank you.

PR #722, opened 2018-12-04
Bug Fix: Special characters in password are not escaped

FelixLC:
https://github.com/sequelize/cli/issues/172 is still happening when special characters find their way in a password

### DIFF
--- a/src/helpers/config-helper.js
+++ b/src/helpers/config-helper.js
@@ -152,7 +152,7 @@ const api = {
   },
 
   filteredUrl (uri, config) {
-    const regExp = new RegExp(':?' + (config.password || '') + '@');
+    const regExp = new RegExp(':?' + _.escapeRegExp(config.password) + '@');
     return uri.replace(regExp, ':*****@');
   },
 


FelixLC:
The test suite is quite scattered around multiple files; I tried to test using docker-compose and init config but I feel it would be better to test `filteredUrl` in  a unit test what do you think ?

sushantdhiman:
@FelixLC unit test works

staadecker:
Any updates? Also see #774

agonza1:
This seems to work. I just tried changing the MySQL password to "12p@ss*word34" and the password was filtered correctly to ***
And there is already a unit test in place: `filters the password` in `test/url.test.js`

tsonev7:
Any idea when this will be pushed to `npm` ?

SimonSchick:
@sushantdhiman

jeremyhon:
This has landed in `v5.5.0`. Thanks @sushantdhiman

pekzanan:
Hello, It seem that there is still a problem when password have : inside
For example, "12p:@ss*word34" it would return 'password authentication failed for user'. I tried to encode the password via URIencode, but not help.

PR #704, opened 2018-10-22
feat(mode:generate) add enum support

mkly:
Add support for `model:generate --name User role:enum:{Admin, Guest}`

Closes #550

PR #728, opened 2018-12-11
Documentation: Update docs for enum type

mkly:
Enum type brackets need to be quoted or escaped in most shells.

* Update documenation and add an addition FAQ item about issue


mkly:
Unfortunately, I failed to account for the shell with the use of brackets. Someone was having difficulty because of that, and I imagine they will not be the only one.

https://github.com/sequelize/cli/pull/704#issuecomment-446071515

I've added some additional clarification to the documentation, but please let me know if you have any other ideas to help alleviate the confusion here. Thank you.

sushantdhiman:
Thanks @mkly

PR #733, opened 2018-12-18
Update README

frank-borkin:
If you npm install sequelize-cli and don't install sequelize, then you get an "Unable to resolve sequelize package in blah" error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ The Sequelize Command Line Interface (CLI)
 Install CLI globally with
 
 ```bash
-$ npm install -g sequelize-cli
+$ npm install -g sequelize sequelize-cli
 ```
 
 Now you can run CLI using following command anywhere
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ The Sequelize Command Line Interface (CLI)
 Install CLI globally with
 
 ```bash
-$ npm install -g sequelize-cli
+$ npm install -g sequelize sequelize-cli
```
sushantdhiman: Globally installed cli should be able to pick locally installed `sequelize`

frank-borkin: Sure, but that's not the point - the point is that a new users (like me) who just want the cli and are getting started will install the cli and it will fail because sequelize isn't installed. So either we update the docs, or we update the deps (which gets more interesting because of the local/global thing).

PR #734, opened 2018-12-23
Fix `config.js` generation

mubaidr:
If `.sequelizerc` has `config` file specified as '.js' file. Running `init` will generate invalid config file (with json content).

This patch fix this issue.


SimonSchick:
Can you add a tiny test for this?

mubaidr:
Running tests fails with error :(

```bash
Uncaught Error: Command failed: D:\current\cli\lib\sequelize init
'D:\current\cli\lib\sequelize' is not recognized as an internal or external command,
operable program or batch file.
```

all dependencies have been installed.

PR #735, opened 2018-12-26
Add async to migration skeleton functions

townofdon:
This is a minor enhancement and would speed up workflow for those using `async/await`.

All tests passing, except for the following which seems unrelated to my changes:

```
290 passing (9m)
  1 failing

  1) [SQLITE] lib/sequelize model:create
       attributes
         --attributes first_name:string,last_name:string,bio:text,role:enum:{Admin,"Guest User"},reviews:array:text
           exits with exit code 0:
     Uncaught Error: Command failed: /Users/don/websites/js/sequelize/cli/lib/sequelize model:create --name User --attributes first_name:string,last_name:string,bio:text,role:enum:{Admin,"Guest User"},reviews:array:text
ERROR: Attribute 'role:enum:Admin' cannot be parsed: Unknown type 'Admin'

      at ChildProcess.exithandler (child_process.js:288:12)
      at ChildProcess.EventEmitter.emit (domain.js:442:20)
      at maybeClose (internal/child_process.js:962:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:249:5)
```

### DIFF
--- a/src/assets/migrations/skeleton.js
+++ b/src/assets/migrations/skeleton.js
@@ -25,5 +25,5 @@ module.exports = {
       Or, await execution:
       queryInterface.dropTable('users');
     */
-  }
+  },
 };
--- a/src/assets/migrations/skeleton.js
+++ b/src/assets/migrations/skeleton.js
@@ -1,23 +1,29 @@
 'use strict';
 
 module.exports = {
-  up: (queryInterface, Sequelize) => {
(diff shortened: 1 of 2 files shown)


SimonSchick:
Not every code base uses node v8 yet and this would be extra churn, not merging before we go v8+

eseliger:
what else would work is adding a `process.version` check, if we want to proceed with this more quickly

PR #741, opened 2019-01-29
Show original error details

kroleg:
This patch is very useful for debugging errors which happens when you run db:migrate.
Error before (not so user-friendly):
```
ERROR: Validation error
```
Error after (much easier to find what causes it, right?):
```
ERROR: Validation error
Message: could not create unique index "unique_slug". Details: Key (slug)=(dota2-pro-league-1-open-qualifiers) is duplicated.
```

### DIFF
--- a/src/helpers/view-helper.js
+++ b/src/helpers/view-helper.js
@@ -24,22 +24,22 @@ module.exports = {
 
   error (error) {
     let message = error;
-    let extraMessages = [];
+    const extraMessages = [];
 
     if (error instanceof Error) {
       message = !args.debug
         ? error.message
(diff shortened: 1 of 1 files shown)


jackycute:
I had same issue, could this been merged soon?

ff-will:
Oh boy, that would be really helpful!

sushantdhiman:
Fixed by https://github.com/sequelize/cli/commit/ea837e533921bbc5f45702788a8cc95df50009dc

sushantdhiman:
Extra error message will be printed with `--debug` flag

PR #747, opened 2019-02-25
Recover from errors when running migrations in a non-default schema with a user who cannot create schemas

pettyalex:
My team encountered this and had to revert to an older version of sequelize-cli in order to keep moving forward.

Our migrations are run as a user that does not have permissions to create new schemas, but has full access to the custom schema passed in. When sequelize-cli started attempting to create the custom schema every time, without checking if the schema exists, we started seeing migrations fail with `ERROR: permission denied for database`. This will give the user some information about what's happening, while letting us move to current sequelize-cli.

### DIFF
--- a/src/core/migrator.js
+++ b/src/core/migrator.js
@@ -34,7 +34,7 @@ function getSequelizeInstance () {
 function createSchemaIfNotExists(sequelize, schemaName) {
   return sequelize.showAllSchemas().then(
     schemaNames => {
-      if(schemaNames.indexOf(schemaName === -1)) {
+      if(schemaNames.indexOf(schemaName) === -1) {
         return sequelize.createSchema(schemaName);
       }
       return;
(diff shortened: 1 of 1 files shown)


pettyalex:
@sushantdhiman This would help out my team, can we get a review / merge? Thank you.

src/core/migrator.js
```diff
@@ -67,7 +68,13 @@ export function getMigrator (type, args) {
         if (helpers.version.getDialectName() === 'pg') {
           const customSchemaName = helpers.umzug.getSchema('migration');
           if (customSchemaName && customSchemaName !== 'public') {
-            return sequelize.createSchema(customSchemaName);
+            return sequelize.createSchema(customSchemaName).catch(err => {
```
sushantdhiman: This can be changed to only create schema if it doesn't exist, [`showAllSchemas`](https://github.com/sequelize/sequelize/blob/d43f162d75203ddadfc4a2479215d53204cc5c7b/lib/query-interface.js#L111) can be used

pettyalex:
@sushantdhiman This bit us again, so I corrected it to use `showAllSchemas` and only attempt to create the schema if it does not already exist

pettyalex:
This issue is still present.

sushantdhiman:
Needs rebase

PR #788, opened 2019-05-31
propagating ssl=true from --url into dialectOptions

igorski89:
When connecting to the database using the `--url` option I noticed the ssl options are ignored. The only option being to revert to a config file specifying `dialectOptions` directly there.

While it worked just fine it felt too cumbersome to manage when the rest of the app configuration is made via env variables.

Digging a little bit into the code turns out the url parser fully supports the `ssl=true` option yet doesn't propagates it further into the `config` object to have a Sequelize instance initialised.

This small change addresses the issue.

Fixes https://github.com/sequelize/cli/issues/154

### DIFF
--- a/src/helpers/config-helper.js
+++ b/src/helpers/config-helper.js
@@ -187,7 +187,10 @@ const api = {
     let config = api.urlStringToConfigHash(urlString);
 
     config = _.assign(config, {
-      dialect: config.protocol
+      dialect: config.protocol,
+      dialectOptions: {
+        ssl: config.ssl
+      }
     });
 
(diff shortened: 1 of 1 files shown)


ahimta:
This is especially important because you will most likely have all users with a require-ssl option and wouldn't be able to connect otherwise:sweat_smile:.

However, the bigger issue maybe that many stuff in the URL are not passed and not just SSL:sweat_smile:.

mannol:
This would be a really great feature to have. It seems straight forward enough for merge, @sushantdhiman?

dmayle:
@mannol @sushantdhiman @papb This actually fixes a real issue, any chance it can still be merged?

PR #822, opened 2019-09-07
migrations and seeds with both raw sql and js interface

LucasBadico:
we need to run migrations with raw sql scripts as I show in #820, but we still want to be able to run the original migrations from sequelize, like those bellow:
Captura de Tela 2019-09-06 às 17 47 15

for this I will add a option on `migration:create --name migrationName --raw` and `seed:create  --name seedName --raw` for creating those 3 files. The sql one is empty, and the js will have this content:
```js
    const { run } = require('sequelize-cli');
    module.exports = run(__filename, __dirname);
```
This runner will read the sql files with same name as the `__filename` and build the migration interface of sequelize.

### NOTE: this runner is already working.

I think that this aprouch is better because we will not change the runtime of migration, a lot less sideeffect. The only one found until now is a console.log message saying that the sql files do not macth the regex for migration file, but as we are using the js file to read those, its not have any importance.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "sequelize": "./lib/sequelize",
     "sequelize-cli": "./lib/sequelize"
   },
+  "main": "./lib/index",
   "dependencies": {
     "bluebird": "^3.5.3",
     "cli-color": "^1.4.0",
--- a/src/commands/migration_generate.js
+++ b/src/commands/migration_generate.js
@@ -19,18 +19,28 @@ exports.builder =
 exports.handler = function (args) {
   helpers.init.createMigrationsFolder();
 
-  fs.writeFileSync(
-    helpers.path.getMigrationPath(args.name),
-    helpers.template.render('migrations/skeleton.js', {}, {
-      beautify: false
-    })
-  );
-
-  helpers.view.log(
-    'New migration was created at',
-    clc.blueBright(helpers.path.getMigrationPath(args.name)),
-    '.'
-  );
-
+  if (args.raw) {
+    helpers.raw.write(helpers.getPath('migration'), args.name);
+    helpers.view.log(
+      'New migration files was created at',
+      clc.blueBright(helpers.path.getPath('migration')),
+      '.'
+    );
+  
+  } else {
+    fs.writeFileSync(
+      helpers.path.getMigrationPath(args.name),
+      helpers.template.render('migrations/skeleton.js', {}, {
+        beautify: false
+      })
+    );
+  
+    helpers.view.log(
+      'New migration was created at',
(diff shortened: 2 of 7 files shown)


papb:
Hello, thanks for taking the time to help! Is your goal to have a JS migration that just runs the SQL queries? Why do we need this extra "boilerplate" file? Wouldn't it be better to simply run the sql files?

LucasBadico:
> Hello, thanks for taking the time to help! Is your goal to have a JS migration that just runs the SQL queries? Why do we need this extra "boilerplate" file? Wouldn't it be better to simply run the sql files?

Yes, my first attemp was to do the migrator read the sql file, but, I was forcing a lot of the logic into the migrator to know when to call `.up.sql` or `.down.sql`.

In this way, for the migrator all migrations still be js. So is transparent.

papb:
Where is the `sequelize-cli/lib/sql-runner` file?

LucasBadico:
> Where is the `sequelize-cli/lib/sql-runner` file?

WHOOPS,  I copy the old code, the actual code is this:
```
 const { run } = require('sequelize-cli');
 module.exports = run(__filename, __dirname);
```

LucasBadico:
Just pushed the last iteration. Let me ask something, about the tests, I have 3 failing test commig from master. Do you guys know about them? How can I fix them and make my own tests for this feature.

papb:
Looks like you figured it out already, they were linting errors, right?

I will take a look on your PR later, I am a bit busy right now

LucasBadico:
> Looks like you figured it out already, they were linting errors, right?
> 
> I will take a look on your PR later, I am a bit busy right now

No, not quite, the travis ci is just looking for lint errors, it is not running the suite of tests. So I just fixed the lint errors and the travis ci is in green. But still failing on the three tests. I'm exploring this errors right now.

LucasBadico:
I found the failing tests, skipped them to finish my PR.

papb:
> No, not quite, the travis ci is just looking for lint errors, it is not running the suite of tests.

This is not true, the travis CI is running everything. When you had linting errors, it stopped early, but now that everything is passing, everything is run.

Since the tests were failing only on your machine and not on Travis CI, perhaps it was a language issue? The test is checking for a string match in english, perhaps your tests got some portuguese error messages...

Please undo your last commit, we don't want to skip tests

LucasBadico:
I figurout this after I push the commits, will undo, sorry about that.

LucasBadico:
Reverted, I will start the tests this night.

LucasBadico:
Its better you take a look after I finish the tests. I give here a headsup to you.

PR #824, opened 2019-09-11
Add support of file extensions for .sequelizerc config

octet-stream:
This pull request improves .sequelizerc config loading by adding an ability to use extensions in the name. Now CLI will use [`cosmiconfig`](https://github.com/davidtheclark/cosmiconfig), so it will try load config from (according to cosmiconfig documentation):
- a `sequelize` property from package.json file
- `.sequelizerc` file
- `.sequelizerc.json` file
- `.sequelizerc.yml`, `.sequelizerc.yaml` or `.sequelizerc.js` files
- `sequelize.config.js` file

Since I have an obstacle with tests, I have decided to open this pull request to get your help with that.
Tests failing with the following message, so I can't be sure if everything is okay and the new feature passes:
<details>
<summary>Failing tests</summary>

iTerm2_2019-09-12 02-29-48@2x

</details>

I'm using Node 12.9.1, Yarn 1.17.3 and macOS 10.14.6 to run tests.

### DIFF
--- a/src/core/yargs.js
+++ b/src/core/yargs.js
@@ -1,13 +1,13 @@
-import fs from 'fs';
 import yargs from 'yargs';
 import path from 'path';
+import cosmiconfig from 'cosmiconfig';
 
-function loadRCFile(optionsPath) {
-  const rcFile = optionsPath || path.resolve(process.cwd(), '.sequelizerc');
-  const rcFileResolved = path.resolve(rcFile);
-  return fs.existsSync(rcFileResolved)
-    ? JSON.parse(JSON.stringify(require(rcFileResolved)))
-    : {};
+const explorer = cosmiconfig('sequelize', {stopDir: process.cwd()});
+
+function loadRCFile (optionsPath) {
+  optionsPath = path.resolve(optionsPath || process.cwd());
+
+  return explorer.searchSync(optionsPath).config;
 }
 
 const args = yargs
--- a/src/core/yargs.js
+++ b/src/core/yargs.js
@@ -2,7 +2,20 @@ import yargs from 'yargs';
 import path from 'path';
 import cosmiconfig from 'cosmiconfig';
 
-const explorer = cosmiconfig('sequelize', {stopDir: process.cwd()});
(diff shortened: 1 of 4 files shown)


octet-stream:
The implementation is done, but I still can't pass through tests, so I open PR for a review and hopefully get some help with that.

frlinw:
It wont be possible to have a 'no extension' file with js inside with Node 13.x (because esm module requires an extension)
It would be great to review this PR or `--option-path` will soon become required.

octet-stream:
Well I do think is that config loading must follow default cosmiconfig behaviour to load files without files extension. I did a custom loader just for backward compatibility with the current major version of sequelize-cli.

src/core/yargs.js
```diff
@@ -1,13 +1,28 @@
...
-    : {};
+function noExtLoader (filename, content) {
+  try {
+    return cosmiconfig.loadJson(filename, content);
+  } catch (_) {
+    return require(filename);
+  }
+}
+
+const explorer = cosmiconfig('sequelize', {
```
frlinw: If you plan to update to cosmiconfig v6, the API has changed. You should now use `cosmiconfigSync()` and `explorer.search()`

octet-stream: Sure, if they will drop Node < 8 support (currently the minimal version for CLI is 6 according to package.json) since Node 8 is minimal version fro cosmiconfig.

papb: Hi everyone, I am currently super busy maintaining the main repo, that's why I haven't shown up around in the CLI repo, thanks for the patience (and by the way I don't think I will be showing up in the near future as well), just thought I'd say that

> if they will drop Node < 8 support

we definitely will do that... Probably drop Node < 10 actually.

octet-stream: It's okay, I bet you're currently working on v6 and resolving of tones of issues in the main repo :)

> we definitely will do that... Probably drop Node < 10 actually.

Nice, then I'll move to the next major version of cosmiconfig as well, once you'll drop end-of-life Node versions.

PR #844, opened 2019-10-26
Prevent error on db:drop if database does not exist

### LINKED ISSUE #840: db:drop should not fail if not exists
## What you are doing?
when i run the command sequelize db:drop and the database does not exist, i get an error.

the command to be ignored as there was no database to delete

## Issue insight
The error is due to the command being used without the "IF EXISTS" clause, despite this clause being common to all providers who support db:drop on the first place Mysql, Postgres and MSSql

to fix this we only need to change the /src/commands/database.js from this
```js
 await sequelize.query(`DROP DATABASE ${sequelize.getQueryInterface().quoteIdentifier(config.database)}`, {
        type: sequelize.QueryTypes.RAW
      }).catch(e => helpers.view.error(e));

```js

to this 
```js
 await sequelize.query(`DROP DATABASE IF EXISTS ${sequelize.getQueryInterface().quoteIdentifier(config.database)}`, {
        type: sequelize.QueryTypes.RAW
      }).catch(e => helpers.view.error(e));
```

glhrmv:
Closes #840.

### DIFF
--- a/src/commands/database.js
+++ b/src/commands/database.js
@@ -65,7 +65,7 @@ exports.handler = async function (args) {
 
       break;
     case 'db:drop':
-      await sequelize.query(`DROP DATABASE ${sequelize.getQueryInterface().quoteIdentifier(config.database)}`, {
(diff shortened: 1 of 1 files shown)


papb:
Thanks!

PR #850, opened 2019-11-19
add filename-date-format flag

ryuichi7:
Hello! This is a request to add a new flag, `--filename-date-format` to the `migration:create` command.

This would allow for two new formats:
`unix-timestamp`
`unix-timestamp-millis`

so that this command:
`npx sequelize-cli migration:create --name foo --filename-date-format unix-timestamp-millis`
would create a file named:
`1574156777964-foo.js`

This change is fully backwards compatible and defaults to the `YYYYMMDDHHmms` date format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 
 ## Future
 
+### Feature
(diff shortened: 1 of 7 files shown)


sushantdhiman:
I don't think there will be any advantage of any other date format. We just need to sort files, which is handled properly with current setup

PR #857, opened 2019-12-16
feat(chore): stick with yargs API

jamesgeorge007:
Stick with the `yargs` API to show up by default when no commands were passed in.


sushantdhiman:
Fixed in master

Issue #578, opened 2017-10-25
No possible way to get table name to singular


ajbeach2:
There is no current way to override this method:
https://github.com/sequelize/cli/blob/master/src/helpers/migration-helper.js#L8

I need to use singular table names because loopback doesn't follow conventions of having plural table names. I am using sequelize so I don't have to run auto migrations in loopback which are problematic.

How difficult would it be to have a "freezeTableName" configuration somewhere to allow for singular table names?

PR #869, opened 2020-02-18
freezeTableName support

kiro112:
Add support to allow singular table name.

Note: I cant think of name that can replace `_underscoreOption`.

```
sequelize-cli model:generate --underscored --name user --attributes first_name:string,last_name:string,email:string```


kiro112:
#578

PR #883, opened 2020-04-23
Remove operatorsAliases from the default config

negreirosleo:
There's no point in having operatorAliases at the default config since [it is deprecated in sequelize v5](https://sequelize.org/master/manual/model-querying-basics.html#deprecated--operator-aliases) 

Manually removing it after `sequelize init` is an unnecessary step that we can cut to provide a better developer experience :)
image


sushantdhiman:
Fixed in v6 version

PR #899, opened 2020-05-30
Add support for timestamps in migration tables

ShaharHD:
The relevant fields are: `seedersTimestamps` and `migrationTimestamps` which can be set in the sequelize `config.json`

The code will also autodetect if the table already has the timestamps and will act upon it.

### DIFF
--- a/src/helpers/umzug-helper.js
+++ b/src/helpers/umzug-helper.js
@@ -15,6 +15,8 @@ const storageJsonName = {
   seeder: 'sequelize-data.json',
 };
 
+let timestampsDefault = false;
+
 module.exports = {
   getStorageOption(property, fallback) {
     return helpers.config.readConfig()[property] || fallback;
@@ -41,6 +43,14 @@ module.exports = {
     return this.getStorageOption(type + 'StorageTableSchema', undefined);
   },
 
+  enableTimestamps() {
+    timestampsDefault = true;
+  },
+
+  getTimestamps(type) {
+    return this.getStorageOption(type + 'Timestamps', timestampsDefault);
+  },
+
   getStorageOptions(type, extraOptions) {
     const options = {};
 
@@ -49,6 +59,7 @@ module.exports = {
     } else if (this.getStorage(type) === 'sequelize') {
       options.tableName = this.getTableName(type);
       options.schema = this.getSchema(type);
+      options.timestamps = this.getTimestamps(type);
     }
 
     _.assign(options, extraOptions);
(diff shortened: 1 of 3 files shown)


ShaharHD:
@sushantdhiman trying to get your attention here as well (as you've been active on sequelize)

Can you please review this? or point me out to who can review this?

Thanks

sushantdhiman:
This needs test

src/helpers/umzug-helper.js
```diff
@@ -15,6 +15,8 @@ const storageJsonName = {
   seeder: 'sequelize-data.json'
 };
 
+let timestampsDefault = false;
```
sushantdhiman, line 18: I don't like using this out of scope variable to track timestamp default

ShaharHD: Open to suggestion - how else would you want to track this? I wanted to keep this a limited module variable - but it can be modified to be as a variable in the `module.exports`

ShaharHD:
I will try and revise a test for this.

src/helpers/umzug-helper.js
ShaharHD: @sushantdhiman ?

ShaharHD:
> This needs test

@sushantdhiman Added a test and updated the PR

Look for `run migration again with timestamp fields present` without the PR fixes it will fail, as it will not feed the relevant fields of timestamps when running new migrations.

If I need to fix it for v5 as well, which branch should I push it to?

src/helpers/umzug-helper.js
sushantdhiman: I don't see any good way either, we can keep this

sushantdhiman:
With CLI we don't port back to earlier releases, so no need for v5 PR

ShaharHD:
> With CLI we don't port back to earlier releases, so no need for v5 PR

@sushantdhiman I could only find an outdated `5.21.11` greenkeeper branch on the cli repository.

I have created [v5](https://github.com/ShaharHD/sequelize-cli/tree/v5) on which I based on https://github.com/sequelize/cli/commit/d4549d4bf360e87f2b0a588dbafa3b76ea7a6230 

To which branch should I point the PR to? there's no v5 on this repo

sushantdhiman:
> Because by default npm install sequelize-cli will install it as "sequelize-cli": "^5.5.1" so unless the new version will be in the 5.x.x range, it will not pick it up on new installs.

That is how npm install works.

If you are already using v5/v6 of sequelize, you should be using latest release of the CLI. There are no major breaking changes apart from Node version update

PR #915, opened 2020-07-05
Support migrations files with ts file extension

kroleg:
### Pull Request check-list

### Description of change

Support *.ts migration files

### DIFF
--- a/src/core/migrator.js
+++ b/src/core/migrator.js
@@ -45,7 +45,7 @@ export async function getMigrator(type, args) {
     migrations: {
       params: [sequelize.getQueryInterface(), Sequelize],
       path: helpers.path.getPath(type),
-      pattern: /\.c?js$/,
+      pattern: /\.(cjs|js|ts)$/,
     },
   });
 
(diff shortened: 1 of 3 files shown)


sushantdhiman:
Released as https://github.com/sequelize/cli/releases/tag/v6.2.0

intellix:
I can't create feature requests without a test case but don't wanna lose the idea so writing here instead.

My current migrations are all JS and would like to transition them to TS migrations so I can import the interfaces to understand the methods available. They're being read but my migrations are starting from scratch and it seems to be because the migration name includes the extension:

```
20181113051813-create-achievement.js
20181113051813-create-achievement.ts
```

I can run an SQL to change the migrated extensions I guess:

```
UPDATE "SequelizeMeta" SET "name" = REPLACE("name", '.js', '.ts')
```

Would be cool if the migration name didn't include the extension (as an option)

mirajehossain:
How do I generate migration in `.ts` extension, please update the doc and give an example to use this feature thanks.

PR #916, opened 2020-07-06
Fixes minor formatting typo in documentation

danajanezic:


papb:
Hello! I see you are a first-time contributor, thank you for taking the time to help Sequelize! However in this case I disagree with the  change, the `d` is left outside intentionally, another possible wording would be:

> `require`-ed

Do you think that would be better?

danajanezic:
That's certainly better, but weird to me. Maybe no backticks at all? I don't know if there's a style convention. Either way, your suggestion is better than what it was before.

papb:
No backticks is good! Please update your PR and thank you!!

PR #924, opened 2020-07-12
Use DataTypes from sequelize instead of injecting into the migrations and Models

Celleb:
Import and use DataTypes from `sequelize` instead of injecting into the migrations and models

### Pull Request check-list

### Description of change

Import and using DataTypes from `sequelize` allows for auto completion and type hinting.


Celleb:
Unfortunately, there are tests failing on the current master branch.

PR #934, opened 2020-12-16
Add db:create and db:drop support for MariaDB

pbstriker38:
### Pull Request check-list

**No, but linting issues are unrelated to my change.**
**I have db drop and create tests that pass. Some of the other pre-existing tests do not pass locally for me.**

### Description of change

Adds support for db:create and db:drop for MariaDB. This uses the same code block as MySQL since they are compatible.

I can push up the tests if you want. It's just an extra `if (Support.dialectIsMariaDB()) {` in the db-create and db-drop test files.

### DIFF
--- a/src/commands/database.js
+++ b/src/commands/database.js
@@ -109,6 +109,7 @@ function getCreateDatabaseQuery(sequelize, config, options) {
       );
 
     case 'mysql':
+    case 'mariadb':
       return (
         'CREATE DATABASE IF NOT EXISTS ' +
         queryGenerator.quoteIdentifier(config.database) +
@@ -163,6 +164,7 @@ function getDatabaseLessSequelize() {
       break;
 
     case 'mysql':
+    case 'mariadb':
       delete config.database;
       break;
 


WikiRik:
Hi, thanks for submitting this PR. And our apologies for the late response. Are you still willing to add some tests to this?

msanguineti:
I was trying to fix this problem but I have found inconsistencies in how `sequelize` returns the list of tables for a database schema. These inconsistencies cause failures in tests which then needs special handling making them a bit brittle.

`sequelize` returns an array of strings with table names for `mysql`, `pgsql` and `sqlite` while it returns an array of objects for `mariadb` and `mssql`.

```js
// `mysql`, `pgsql` and `sqlite`
showTableNames() -> ['tableA', 'tableB', ...]

// `mariadb` and `mssql`
showTableNames() -> [{tableName: 'tableA', schema: 'schemaX}, {tableName: 'tableB', schema: 'scehmaX}, ...]
```

So, I could write tests to handle this inconsistencies thus making the `cli` handle `create/drop` for `mariadb`, but I am wondering if this issue should really be fixed in `sequelize` itself by normalising the API across all the dialects. Also, I am wondering how tests for `mssql` are passing. I do not have a MSSQL server I can test against so my doubts remain.

WikiRik:
> I was trying to fix this problem but I have found inconsistencies in how `sequelize` returns the list of tables for a database schema. These inconsistencies cause failures in tests which then needs special handling making them a bit brittle.
> 
> `sequelize` returns an array of strings with table names for `mysql`, `pgsql` and `sqlite` while it returns an array of objects for `mariadb` and `mssql`.
> 
> ```js
> // `mysql`, `pgsql` and `sqlite`
> showTableNames() -> ['tableA', 'tableB', ...]
> 
> // `mariadb` and `mssql`
> showTableNames() -> [{tableName: 'tableA', schema: 'schemaX}, {tableName: 'tableB', schema: 'scehmaX}, ...]
> ```
> 
> So, I could write tests to handle this inconsistencies thus making the `cli` handle `create/drop` for `mariadb`, but I am wondering if this issue should really be fixed in `sequelize` itself by normalising the API across all the dialects. Also, I am wondering how tests for `mssql` are passing. I do not have a MSSQL server I can test against so my doubts remain.
> 
> EDIT:
> 
> I have tried to run the tests against a MSSQL database using the [docker image](https://hub.docker.com/_/microsoft-mssql-server) provided by Microsoft and indeed the tests are failing due to the same reason that the returned list of table names is a list of objects and not a simple list of strings (names).

Hi. Thanks for looking into this!
It would indeed be good if it is normalized in sequelize itself, but if a dialect does not support schemas it would be prettier to return only the table names. So I don't think a fix like this would be the best solution. I think we should just accommodate for both options. I will work on update the CI to include the same dialects as sequelize itself and then maybe we can work together on fixing the failing tests?

ZeroExistence:
Hi, just want to check if there will be an update on reviewing this `mariadb` support for create/drop action. Thank you.

boadadf:
For God's sake, someone please, accept this PR.
It's just a new case to the switch. It can't break anything since the syntax for mysql and mariadb are the same!

PR #936, opened 2021-01-25
If using createdAt/updatedAt default to now

mrgcohen:
### Description of change

This defaults createdAt/updatedAt to now.


WikiRik:
@sdepold not sure if there is a reason that this was not implemented earlier, but seems like a good thing to add. Do you agree?

Americas:
Sequelize itself already handles timestamps, doing this is not wrong, but it sort of assumes that you know what you are doing, having this by default could lead to wrong assumptions.

And where would it stop? Should the migration also add a trigger to update the `updatedAt` attribute on row updates?

Issue #157, opened 2015-07-21
Generate initial migration from model?


jsumners:
I'm just getting started with Sequelize and the CLI. I want my project's initial database to be bootstrapable with `sequelize db:migrate` (i.e. create the database from nothing with an initial set of migrations). In that regard, it seems to me that I am duplicating a lot of code.

1. I create my initial model, wherein I fully describe the table
2. I then create a migration for it where I mostly just copy and paste the previously written definition
3. If there are changes to be made before committing, I get to make them in multiple places

I have a simple two table test project at https://github.com/jsumners/sequelize-migration-test that shows what I'm describing.

Would it not be possible (make more sense) for `sequelize migration:generate` to read the already defined model and generate the migration from that definition?

sdepold:
Yes. That would be a good idea and I'm planning to do that since a while already but never found the time for that :(

jsumners:
If I get a chance I'll see if I can work something up. But I'm not sure I'll have the opportunity.

o5:
+1 :)

mwain:
+1

DenLilleMand:
+1

btroo:
+1

golendukhin:
+1

amit034:
Is there any progress? or alternative?

sdepold:
no progress afaik

zmts:
+1 Will be the very popular feature. I hope this feature will be implemented

kkleidal:
+1

suricactus:
+1

oliv23:
+1

WaqqasMeraj:
+1

juniorplenty:
+1

atlantageek:
+1

bradleyy1012:
+1

jcfinnerup:
+1

Anomnimel:
+1

darrenchiu:
+1

zmts:
By the way, now I use https://github.com/Vincit/objection.js/ 
Very nice convenient ORM

notgiorgi:
+1

Though currently I'm doing it this way: https://gist.github.com/notgiorgi/7276aae5baebb5725799ce0f5b8468f3

SargeKhan:
I would like to start working on this feature. Can anyone give me pointer on where I should begin?
Thanks alot.

omeragich:
+1

iDVB:
+1

[...]

PR #942, opened 2021-04-06
Fix/#157

hrbocutti:
Issue: #157 
Desc: Adding a verification file if has extension TS or JS to migrate with skeleton


WikiRik:
Hi, thanks for submitting a PR! I'm unfortunately not sure how this would fix #157, especially since it seems that [sequelize-auto](https://github.com/sequelize/sequelize-auto) might provide this feature
Could you expand a bit on what problems this PR is solving?

hrbocutti:
Thanks @WikiRik Maybe this solution are not applicable now.
I'm close this PR.

Maybe can I could help with another PR, for another issue.

Bye!

PR #972, opened 2021-11-04
Feature/ci

sdepold:
This is adding releasing with semantic-release

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [10, 12, 14]
+        sequelize-version: [5, latest]
     name: Postgres (Node ${{ matrix.node-version }})
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -87,5 +90,34 @@
   "homepage": "https://github.com/sequelize/cli",
   "engines": {
     "node": ">=10.0.0"
+  },
+  "commitlint": {
+    "extends": [
+      "@commitlint/config-angular"
```
WikiRik, line 96: Any specific reason why you use this configuration?

package.json
```diff
@@ -87,5 +90,34 @@
...
+          "revert",
+          "style",
+          "test",
+          "meta"
+        ]
+      ]
+    }
+  },
+  "husky": {
+    "hooks": {
```
WikiRik, line 119: Maybe add a pre-commit hook that runs the linter?

.github/workflows/ci.yml
```diff
@@ -8,6 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [10, 12, 14]
```
WikiRik, line 10: Node 16 is now LTS, maybe include that as well? Not sure what the policy is from the main repo on supported Node versions

WikiRik (commented):
Looks good, just have a few questions/remarks

package.json
WikiRik: And possibly with lint-staged to make it quicker

sdepold:
@WikiRik I'll have a look into your suggestions today or throughout the weekend!

package.json
sdepold: it's a copy from sequelize/sequelize

PR #974, opened 2021-11-06
Feature/improve ci

sdepold:
Follow-up to #972

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10, 12, 14]
+        node-version: [10, 12, 14, 16]
         sequelize-version: [5, latest]
     name: Postgres (Node ${{ matrix.node-version }}, Sequelize ${{ matrix.sequelize-version }})
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


WikiRik (approved):
LGTM
I do notice that it runs twice now. One time because of a new push and one time because you created the PR. Not sure if that's something that we can prevent safely though

sdepold:
I'm confused too but I think that's somehow normal?!

WikiRik:
I'm not skilled in GitHub Actions, but looking at the documentation it seems that we can use something like this in line 3 (since we do not push directly to main);
```
on:
  pull_request:
    types: [opened, synchronize, reopened]
```

Documentation says about synchronize; 

> Triggered when a pull request's head branch is updated. For example, when the head branch is updated from the base branch, when new commits are pushed to the head branch, or when the base branch is changed.

But I just copied it from the sequelize/sequelize repo. Maybe that people there had a reason to do it this way.


https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

Americas:
You can define the action to run only on pushes to the main branch, and on pull requests. Probably like:

```
on:
  pull_request:
  push:
    branches:
      - 'main'
```

sdepold:
Funny how now one seems to care that it's all broken on node 16 :P

Americas:
> Funny how no one seems to care that it's all broken on node 16 :P

Not all, one passed 😄

WikiRik:
Tests run so we should be good to merge right?

PR #976, opened 2021-11-06
style: update eslint and prettier

WikiRik:
Just updating eslint, prettier and related devDependencies. Also fixes an occurance of https://eslint.org/docs/rules/no-loss-of-precision by converting it to a bigint as suggested by VSCode


WikiRik:
Well, I forgot that ESLint 8 does not support Node 10 anymore. Any plans on dropping Node 10 support? It reached end of life on May 1st, 2021.

Either way, I'll close this one until then

PR #977, opened 2021-11-08
meta(mocha): move mocha config to own file

WikiRik:
In some of the Node 16 tests for #974 it seemed that mocha was not on a timeout of 30000 but on the default 2000. Not sure why but by moving it to a separate config file it works [fine](https://github.com/WikiRik/cli/actions/runs/1435493979). The config can also be a part of package.json as a mocha property but I prefer to move it to its own config file.

Reporter spec was removed since it is the [default](https://mochajs.org/#-reporter-name-r-name).

I also removed the script, since it only required one thing so it can also be a part of the config file. As per; https://github.com/mochajs/mocha/blob/master/example/config/.mocharc.yml

### DIFF
--- a/.mocharc.yaml
+++ b/.mocharc.yaml
@@ -1,0 +1,7 @@
+bail: true
+check-leaks: true
+color: true
+exit: true
+recursive: true
+require: '@babel/register'
+timeout: '30000'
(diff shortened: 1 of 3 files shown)


WikiRik:
@sdepold this should fix the Node 16 issues, or at least it did on my fork. We'll find out soon enough if that also works here

PR #978, opened 2021-11-09
feat(types): Improve IntelliSense when generating migration files

dtinth:
### Pull Request check-list

- [ ] ~~Have you added new tests to prevent regressions?~~
- [ ] ~~Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?~~

### Description of change

Problem: No IntelliSense when working on migration files

image

Solution: Add a type definition file and use `/** @type */` tag to make editors infer the type of each variable.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "sequelize-cli",
   "version": "6.3.0",
   "description": "The Sequelize CLI",
+  "types": "./types.d.ts",
   "bin": {
     "sequelize": "./lib/sequelize",
     "sequelize-cli": "./lib/sequelize"
--- a/src/assets/migrations/create-table.js
+++ b/src/assets/migrations/create-table.js
@@ -1,5 +1,6 @@
 'use strict';
 
+/** @type {import('sequelize-cli').Migration} */
 module.exports = {
   up: async (queryInterface, Sequelize) => {
     await queryInterface.createTable('<%= tableName %>', {
--- a/src/assets/migrations/skeleton.js
+++ b/src/assets/migrations/skeleton.js
@@ -1,5 +1,6 @@
 'use strict';
 
+/** @type {import('sequelize-cli').Migration} */
 module.exports = {
   up: async (queryInterface, Sequelize) => {
     /**
--- a/src/assets/seeders/skeleton.js
+++ b/src/assets/seeders/skeleton.js
@@ -1,5 +1,6 @@
 'use strict';
 
+/** @type {import('sequelize-cli').Migration} */
 module.exports = {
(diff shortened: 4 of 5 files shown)


hotrungnhan:
hey i working on it , but mine i thought it better a bit.

hotrungnhan:
hey your type declare is wrong 
The Sequelize has type Datatypes , it not sequelize object

dtinth:
> hey your type declare is accidentaly wrong
> The Sequelize has type Datatypes , it not sequelize object

@hotrungnhan Thanks for the suggestions. However, I looked at the [docs](https://sequelize.org/master/manual/migrations.html#migration-skeleton), it says otherwise.

image

The code here on line 46 also says that the migration will be given the full `Sequelize` object; not just the `DataTypes`.

https://github.com/sequelize/cli/blob/54ec29d65727873f88d0daa50e5e92ef7d00a4b8/src/core/migrator.js#L41-L50

Therefore I believe my typings are already correct.

WikiRik:
Hi, thanks for your contribution to sequelize-cli! Although the docs suggest that the Sequelize object is used for the data types, it might be the case that it also accepts the other things. As you mentioned, we do pass the full object to the migration.
I'm not really knowledgeable on typing, but since this only relates to the typing in the skeleton files I prefer the approach of #981 where the type definitions are done inside the files and require sequelize directly. Could you update this PR to do that? And then I think we can ask a maintainer to look at your PR.

dtinth:
@WikiRik Thank you for your review. 😃 

As a user of the library, I prefer _not_ to add 12 lines of boilerplate comments in each migration file, and the code pattern that I presented in this PR is [the same pattern used in Microsoft's Playwright configuration file](https://playwright.dev/docs/test-configuration):

image

Therefore, I hope the proposed changes can be reconsidered, as it does not affect the runtime semantics of the CLI at all. 🙏

dtinth:
(Meanwhile, I have updated the types file to support more TypeScript versions. The `import type` syntax is quite recent, available in TS 3.8 (newer than DefinitelyTyped support window, 3.7), so users on older TS versions may get an error. Because of this I changed it to another syntax that's available since TS 2.9.)

WikiRik:
Not sure why I assigned @sdepold to this, but I think @ephys would be a better fit. Could you take a look at this and merge at your disposal?

ephys (approved):
Not seeing anything wrong with this. We don't support TS < 4.1 anymore but the changes are good.
Just need to figure out what's going on with the CI

WikiRik:
This is for v6 btw, so support for TS 3.7 is a nice addition

PR #982, opened 2021-11-15
feat: added support for ESM config files

corevo:
### Pull Request check-list

### Description of change

Added support for config files that are ESM, usually the migrations themselves are generated so they are commonjs.  
But on the other hand the config could be in ESM in case the backend uses ESM.  

This PR fixes that by trying to import the config file as an ECMAScript module first, and if it fails it fallbacks to CommonJS.  

I've also had to change the babel config, so that it won't transpile the dynamic import to CommonJS as well.

Let me know if a change to the docs is necessary, and if so where to put it.  

This PR also means that the new tests will fail on Node versions older than 12.20, but the failure will be graceful and CommonJS will work instead, since ESM doesn't work in Node versions earlier than that.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -10,5 +10,13 @@
   ],
   "ignore": [
     "src/assets"
-  ]
+  ],
+  "overrides": [{
+    "test": "./src/helpers/import-helper.js",
+    "presets": [
+      ["@babel/env", {
+        "exclude": ["proposal-dynamic-import"],
+      }]
+    ],
+  }]
 }
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "pg": "latest",
     "pg-hstore": "latest",
     "prettier": "^2.4.1",
+    "semver": "^7.3.5",
     "sequelize": "^6.9.0",
     "sqlite3": "latest",
     "through2": "^4.0.2"
--- a/src/helpers/config-helper.js
+++ b/src/helpers/config-helper.js
@@ -5,6 +5,7 @@ import _ from 'lodash';
 import { promisify } from 'util';
 import helpers from './index';
 import getYArgs from '../core/yargs';
+import importHelper from './import-helper';
 
 const args = getYArgs().argv;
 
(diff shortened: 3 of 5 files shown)


.babelrc
```diff
@@ -10,5 +10,13 @@
   ],
   "ignore": [
     "src/assets"
-  ]
+  ],
+  "overrides": [{
+    "test": "./src/helpers/import-helper.js",
```
WikiRik, line 15: Don't think this one is needed right?

WikiRik:
Hi, thanks for the PR. You can update the docs by making a PR for this file in the main sequelize repo; https://github.com/sequelize/sequelize/blob/main/docs/manual/other-topics/migrations.md

Can you show how it fails on Node <12.20? We will be dropping support for Node 10 soonish, so I'm wondering if we should put this on hold until then or that we can merge this with the next release.

.babelrc
WikiRik: My apologies, I should have rephrased it to be more specific. I'm not very knowledgeable in Babel, so I was wondering what this line would do and was confused with something else. But your explanation is very thorough, thanks for that!

corevo:
I'll create a PR for the docs and mention this PR, do you want me to add code to skip these tests in case the runtime is Node 10?

WikiRik:
> I'll create a PR for the docs and mention this PR, do you want me to add code to skip these tests in case the runtime is Node 10?

You read my mind, I was about to ask that. Would be great if you could do that and then I'll assign a maintainer to look at this PR

corevo:
@WikiRik I didn't know where to put `describeOnlyForESM`, I can move it to `test/support/helpers.js` if you want.

I've also opened https://github.com/sequelize/sequelize/pull/13669 for the doc changes.

sdepold:
I think we can get to this from Wednesday on :) we'll discuss which node versions to support on Tuesday

mariusa:
ping & thanks

ephys:
The current solution is going to swallow errors if the file is esm but throws an error.

Instead of a fallback to require, I'd opt for feature-detecting `import()`: https://stackoverflow.com/questions/60317251/how-to-feature-detect-whether-a-browser-supports-dynamic-es6-module-loading

corevo:
We can try to detect specific ESM errors and re-throw if the module itself errors.

ephys:
`import()` throws `Error: Not supported` but detecting where it comes from would require looking at the stack trace.

We can detect whether `import()` is supported without using eval by `require`ing a file that exports `import('fs')`.

If the require fails, or the exported promise rejected, `import()` is not supported.

mariusa:
This wasn't merged, right?

On a new project using `type=module`, ran `npx sequelize-cli init` and it generated `models/index.js` with
```
module.exports = db;
```

ephys:
It was (through https://github.com/sequelize/cli/pull/987) but it only impacts being able to load esm config files, generation of files hasn't been updated yet

mariusa:
Ah, thanks for clarification

PR #984, opened 2021-11-28
Fix ssl config option

Utsav2:
### Description of change

Fix this error:

```

Loaded configuration file "config/config.json".

ERROR: SSL profile must be an object, instead it's a boolean
```

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -109,7 +109,9 @@ Ensure ssl is specified in both `dialectOptions` and in the base config.
         "dialect":"postgres",
         "ssl": true,
         "dialectOptions": {
-            "ssl": true
+            "ssl": {
+                "require": true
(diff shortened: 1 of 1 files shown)


WikiRik:
I will investigate why this is like this in the README, the current code of cli and the main sequelize repo do not seem to do anything with the SSL profile in the dialectOptions. But maybe it did in v5. 

This is also a nice thing to have automated tests for

WikiRik:
Seems to be related to #997

WikiRik:
It does indeed seem that the requirements for dialectOptions are dependent on the dialect used. I'll close this PR for now but if it's updated to reflect the differences within the various dialects we can re-open this. If somebody else wants to make a new PR for this they're also welcome to do this

PR #987, opened 2022-01-08
feat: add support for ESM config files

ephys:
### Description of change

Continuation of PR https://github.com/sequelize/cli/pull/982

This PR uses feature-detection to detect whether we can use `import()` to import the configuration file.
It also fixes the babel config to properly transpile `import-helper.js` for node 10.

Documentation PR is still https://github.com/sequelize/sequelize/pull/13669

Unit tests fail but I think it's unrelated to this PR. Need to investigate.

Also pinging @corevo :)


WikiRik:
Failing unit test is indeed not your fault, but is something we should look into

WikiRik (approved):
Seems to look good. Will wait a few days for @corevo to respond to this as well. Can you change the title of the PR so we can squash this when merging?

src/helpers/import-helper.js
```diff
@@ -0,0 +1,35 @@
...
+async function importModule(modulePath) {
+  // JSON modules are still behind a flag. Fallback to require for now.
+  // https://nodejs.org/api/esm.html#json-modules
+  if (!modulePath.endsWith('.json') && (await supportsDynamicImport())) {
+    return import(modulePath);
+  }
+
+  // mimics what `import()` would return for
+  // cjs modules.
+  return { default: require(modulePath) };
```
corevo, line 29: This assumes that the module only has a default export, and is not compatible with named exports.

For the purposes of the config that's fine, because the config is exported on the default export when is ESM compatible, but this is not true for all cases.

Which makes me feel like it would be better to invoke `supportsDynamicImport` in `config-helper.js` and either use the import helper, or straight up require.

ephys: I'm not sure I understand what you mean.

If the user is using named exports then the platform is compatible with ESM and `import()` will be used, which supports named exports.

If they're using CJS, there are no named-exports. 

Native `import()` of a CJS file already returns `{ default: module.exports }`.

corevo: I understand what you mean, I think I didn't fully understand what you were saying in #982 

But basically you wanted to write a node 10 compatible `importModule` function, while I was under the impression that you were going to drop support for node 10.

WikiRik: We'll indeed drop Node 10 for v7, but I don't think we want to wait with fixing this until the core sequelize package is ready for that

corevo: Yea, makes sense, I think we can resolve this then.

WikiRik: Great, so that's an approval from you? Then I think we can merge and set up a new release

corevo (approved):

src/helpers/import-helper.js
ephys: Yes exactly :) The point was to replicate `import()` using `require`, the code that uses `import-helper` is then responsible for properly using the module

WikiRik:
@ephys I'll leave the merging up to you

ephys:
Here we go 🎉

src/helpers/import-helper.js
mariusa: > The point was to replicate import() using require

For sequelize v7, will this solution be dropped and using the native `import` be supported instead? Along with `.sequelizerc` no longer having `module.exports` ? Should I open an issue for this?

ephys: Yes, the next major will drop support for node 10 so `import()` will always be used.

Supporting ESM in `.sequelizerc` could already be done in this version though. It should be a small change too.
Simply need to make this line use the new import helper https://github.com/sequelize/cli/blob/54ec29d65727873f88d0daa50e5e92ef7d00a4b8/src/core/yargs.js#L9 + support searching for `.sequelizerc` + `.sequelizerc.json` + `.sequelizerc.cjs` + `.sequelizerc.mjs` + `.sequelizerc.js`.

(feel free to open an issue)

mariusa: Thanks! I've openeded https://github.com/sequelize/cli/issues/989 , which is only for v7.
I think current version is fine as is, would prefer focusing on v7 since sequelize v7 is almost ready.

mariusa:
Can https://sequelize.org/master/manual/migrations.html#dynamic-configuration be used with ESM now that https://github.com/sequelize/cli/pull/987 has been merged?

According to docs, this means
1. `config.js` ESM support (there's no example for config.js with ESM syntax)
2. `.sequelize.rc` telling sequelize to use `config.js` instead of `config.json` (no support yet? Or just not documented?)

ephys:
The documentation PR hasn't been merged yet but it should work (https://github.com/sequelize/sequelize/pull/13669)

mariusa:
Thanks. That PR doesn't have ESM examples for config.js / .sequelize.rc. I don't know how to convert existing format to ESM.

ephys:
`.sequelizerc` itself can't be ESM yet (https://github.com/sequelize/cli/issues/990)

config.js can be ESM the same way other ESM files are handled in Node: 
- either use the .js file extension and set "type": "module" in package.json
- or use the .mjs file extension

and the configuration should be the default export.

I'll update the PR with this comment. Is there any other information you need to add to the doc?

vishnu-buildd:
ESM with sequelize has been a little confusing for me. I tried going through the issues in sequelize and sequelize-cli and my understanding is that we can write ESM code with sequelize but not for migrations using the CLI. The `sequelizerc` file is the limitation? Can we have consolidated information in one page of the documentation. More than happy to help out too.

ephys:
I agree that it's confusing. It's mainly due to us not having enough bandwidth available right now to improve this project as we're hard at work on the main repository

What I'd like to do is rewrite this project as esm and add native support for esm everywhere, as a major release that drops node < 12.

vishnu-buildd:
Sounds good to me. I think every node project should ideally support/adopt ESM as standard. It's the way forward. I mean I rewriting an sdk that uses sequelize at work to ESM modules. That's how I ended up here.

Falinor:
It seems that this PR broke the usage of TypeScript with sequelize-cli 6.4.0.
I think it may be related to the support for ESM which breaks the reading of `config.ts`.

```bash
$ <project>/node_modules/.bin/sequelize db:migrate

Sequelize CLI [Node: 16.14.0, CLI: 6.4.0, ORM: 6.20.1]

ERROR: Error reading "sequelize/config.ts". Error: TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for <project>/sequelize/config.ts
```

PR #988, opened 2022-01-11
ci: update config to reflect core repo

WikiRik:
### Pull Request check-list

### Description of change

Updated the CI config so it is more aligned with the core repo. Do note that v6 and v7 branches do not exist yet, I will make a v6 branch after #987 and this PR have been merged. v7 branch will follow at a later stage.


.github/workflows/ci.yml
```diff
@@ -62,22 +65,32 @@ jobs:
...
-      - run: npm install
+          node-version: 16.x
+      - run: yarn install --frozen-lockfile
       - run: npx semantic-release
+      - id: sequelize
+        uses: sdepold/github-action-get-latest-release@master
+        with:
+          repository: sequelize/cli
+      - run: |
+          curl -XPOST -u "sdepold:${{ secrets.GH_TOKEN }}" -H "Accept: application/vnd.github.v3+json" -H "Content-Type: application/json" https://api.github.com/repos/sequelize/cli/dispatches --data '{"event_type":"Release notifier","client_payload":{"release-id": ${{ steps.sequelize.outputs.id }}}}'
```
WikiRik, line 96: @sdepold two/three questions about this;
1. Is this something we want for CLI as well?
2. Why does it refer to secrets.GH_TOKEN here instead of secrets.GITHUB_TOKEN? (the same is in the core repo, I just copied that from here)
3. If the answer to 1. is positive, the answer to 2. makes us require both; is secrets.GH_TOKEN already set up? If not, could you do that?

sdepold: :wave:

This command is triggering the following flow (we'll have to copy the file over): https://github.com/sequelize/sequelize/blob/main/.github/workflows/notify.yml

The prime reason for this line is:

1. I wanted to decouple the notification part from the release pipeline. 
2. I originally observed published releases in the notification pipeline but it turned out that the pipeline won't trigger if the release happened by some other pipeline in the first place.
3. To circumvent the problem, you can dispatch an event from one pipeline, however that (according to my studies on SO and the web) has to happen from another users "session" (in lack of a better word). So the auth tokens have to differ from the one that is running the original pipeline.
4. In short: The ci.yml pipeline does all the things to ensure that tests are green and that releases happen. It will then send a dispatch event using my (or any other) account to the same repo. In the notify.yml pipeline, we listen for such dispatch events and will notify all the relevant channels (currently Twitter and OpenCollective) about the newly made release.

I think it would be nice to have the notify pipeline in place and I can add the GH_TOKEN secret

sdepold: I just added the secret

WikiRik: I've added the notify pipeline as well, guess you'll need to add those secrets as well

.github/workflows/ci.yml
```diff
@@ -1,13 +1,18 @@
-
 name: CI
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+      - v6
```
sdepold, line 6: Do we need v6 and v7 in this repo?

.github/workflows/ci.yml
```diff
@@ -19,17 +24,16 @@ jobs:
...
       - run: docker-compose up -d ${DIALECT}
       - run: docker run --link ${DIALECT}:db -e CHECK_PORT=${SEQ_PORT::-1} -e CHECK_HOST=db --net cli_default giorgos/takis
-      - run: npm run test
+      - run: yarn test
   test-mysql:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10, 12, 14, 16]
+        node-version: [10, 16]
```
sdepold, line 36: We still aim at supporting v10?

.github/workflows/ci.yml
```diff
@@ -41,17 +45,16 @@ jobs:
...
       - run: docker run --link ${DIALECT}:db -e CHECK_PORT=${SEQ_PORT::-1} -e CHECK_HOST=db --net cli_default giorgos/takis
-      - run: npm run test
+      - run: yarn test
   test-sqlite:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10, 12, 14, 16]
+        node-version: [10, 16]
         sequelize-version: [5, latest]
```
sdepold, line 58: Should we include 7 here?

.github/workflows/ci.yml
sdepold: I think we need the release branch here (instead of 6 and 7)

.github/workflows/ci.yml
```diff
@@ -62,22 +65,32 @@ jobs:
...
     runs-on: ubuntu-latest
-    needs: [test-sqlite, test-postgres, test-mysql]
-    if: github.event_name == 'push' && github.ref == 'refs/heads/release'
+    needs:
+      [
+        test-sqlite,
+        test-postgres,
+        test-mysql
+      ]
+    if: github.event_name == 'push' && (github.ref == 'refs/heads/v6' || github.ref == 'refs/heads/v7')
```
sdepold, line 80: we need to react to pushed to the `release` branch here.

.github/workflows/ci.yml
```diff
@@ -62,22 +65,32 @@ jobs:
...
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
         with:
-          node-version: 12.x
-      - run: npm install
+          node-version: 16.x
+      - run: yarn install --frozen-lockfile
       - run: npx semantic-release
+      - id: sequelize
```
sdepold, line 91: rename to `cli` here?

.github/workflows/ci.yml
```diff
@@ -62,22 +65,32 @@ jobs:
...
+        uses: sdepold/github-action-get-latest-release@master
+        with:
+          repository: sequelize/cli
+      - run: |
+          curl -XPOST -u "sdepold:${{ secrets.GH_TOKEN }}" -H "Accept: application/vnd.github.v3+json" -H "Content-Type: application/json" https://api.github.com/repos/sequelize/cli/dispatches --data '{"event_type":"Release notifier","client_payload":{"release-id": ${{ steps.sequelize.outputs.id }}}}'
```
sdepold, line 96: If we are renaming the step above to cli, then we need to change the steps.sequelize part to steps.cli

sdepold (changes requested):

sdepold:
I just added the OpenCollective and Twitter Secrets

.github/workflows/ci.yml
WikiRik: My idea was that we take this the same way as the core repository. So we're now on v6 (and CLI v6 supports Sequelize v5 and v6) and both go to v7 when dropping Node 10 support and CLI v7 will be focusing on supporting Sequelize v7 (but if we can maintain support for earlier versions of Sequelize that would be nice). Possibly a TypeScript migration for the CLI as well would be nice

.github/workflows/ci.yml
WikiRik: Yes, we follow the same Node versions as core sequelize repo

.github/workflows/ci.yml
WikiRik: If we go ahead with splitting things up in v6 and v7 then no. Otherwise yes

.github/workflows/ci.yml
WikiRik: See reasoning in the comment about the branches above

.github/workflows/ci.yml
WikiRik: I thought this id was internal so not really necessary, but I can change it if you want

.github/workflows/ci.yml
sdepold: Makes sense!

.github/workflows/ci.yml
sdepold: True. Fine with me

WikiRik:
@sdepold I think the comments are resolved, can you approve?

PR #991, opened 2022-01-14
build: add semantic-release config

WikiRik:
Yeah, so in https://github.com/sequelize/cli/pull/972 we didn't notice that it didn't actually add the config for semantic-release. But now it should hopefully work, this is just a copy from the core repo

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,5 +116,23 @@
   },
   "lint-staged": {
     "*.js": "eslint"
+  },
+  "release": {
+    "plugins": [
+      "@semantic-release/commit-analyzer",
+      "@semantic-release/release-notes-generator",
+      "@semantic-release/npm",
+      "@semantic-release/github"
+    ],
+    "branches": [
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -116,5 +116,23 @@
...
+    "branches": [
+      "v6",
+      {
+        "name": "v7",
+        "prerelease": "alpha"
+      }
+    ]
+  },
+  "publishConfig": {
+    "tag": "alpha"
```
sdepold, line 136: I guess when we merge this to the new v6 branch, we'll have to change this to minor as a one-off thing

sdepold (approved):

package.json
sdepold: latest that is

WikiRik: Ah, this is specific for the v7 branch?

WikiRik: I see you've done this already so thanks!

Issue #994, opened 2022-01-14
upgrading from v6.3.0 to v6.4.0 breaks sequelize-cli


MichaelTaylor3D:
Project works fine when migrating with v6.3.0, it breaks when using v6.4.0. I have tried using the `file://` scheme as it asks but it didnt make a difference in the error.

```
Sequelize CLI [Node: 16.0.0, CLI: 6.4.0, ORM: 6.13.0]

ERROR: Error reading "src\config\config.cjs". Error: Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

.sequalizerc
```
var path = require('path');

module.exports = {
  config: path.resolve('src', 'config', 'config.cjs'),
  'migrations-path': path.resolve('migrations'),
  'models-path': path.resolve('src', 'models'),
  'seeders-path': path.resolve('seeders'),
};
```

config.cjs
```
require('dotenv').config();

module.exports = {
  local: {
    dialect: 'sqlite',
    storage: './data.sqlite3',
  },
  test: {
    dialect: 'sqlite',
    storage: './test.sqlite3',
  },
  mirrorTest: {
    dialect: 'sqlite',
    storage: './testMirror.sqlite3',
  },
  mirror: {
    username: process.env.DB_USERNAME,
    password: process.env.DB_PASSWORD,
    database: process.env.DB_NAME,
    host: process.env.DB_HOST,
    dialect: 'mysql',
  },
};
```

ephys:
Have you tried `path.resolve('./src', 'config', 'config.cjs')`?

MichaelTaylor3D:
I just tried it and got the same error
![image](https://user-images.githubusercontent.com/5665004/149591644-d004ac86-31fc-4186-84bc-c4ce592dec4f.png)

This the './src' works as well in v6.3.0
![image](https://user-images.githubusercontent.com/5665004/149591735-1fef3346-d04c-4788-8b68-2936267727df.png)

MichaelTaylor3D:
I added the `./src` but received the same error
![image](https://user-images.githubusercontent.com/5665004/149592042-85e213e2-f1a3-48e0-b250-2979c672b3ef.png)

the suggestion to use `./src` still works on v6.3.0
![image](https://user-images.githubusercontent.com/5665004/149592162-82baa612-b3be-4a6d-a023-894e0295e352.png)

[...]

PR #995, opened 2022-01-15
fix: fix config import breaking on windows

ephys:
### Pull Request check-list

- [ ] Have you added new tests to prevent regressions? **Can't as we don't test on windows**

### Description of change

https://github.com/sequelize/cli/pull/987 introduced a bug that caused sequelize-cli to be unusable on windows, as it tries to import the config file using `import()` instead of `require()`. Unlike `require`, `import` expects the path to be a valid URL.

Closes https://github.com/sequelize/cli/issues/994

### DIFF
--- a/src/helpers/import-helper.js
+++ b/src/helpers/import-helper.js
@@ -1,3 +1,5 @@
+const url = require('url');
+
 async function supportsDynamicImport() {
   try {
     // imports are cached.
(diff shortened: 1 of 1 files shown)


WikiRik:
Is there some way we can add a regression test for this?

ephys:
I wish, but unless we start testing on windows, even trying to replicate a windows path will not work :/

image

WikiRik (approved):

WikiRik:
Then we'll just have to accept this for now

PR #996, opened 2022-01-20
build: add codeql analysis

WikiRik:
This PR adds [CodeQL analysis](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql). Config file was generated automatically, but I did remove that it runs on all pushes to main and on a schedule since running it with every PR to main is enough in my opinion.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,66 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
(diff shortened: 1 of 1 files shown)


WikiRik:
Not sure why I drafted this before. Should be good enough to add. We'll fix the issues that come up in different PRs.
If this works well I'll add this to the main repo as well

ephys (approved):
This looks good to me. But I think we can remove the `******** NOTE ********` part of the workflow file

WikiRik:
Good point, I've removed all comments since we can easily find them back in the Actions template

ephys (approved):

ephys:
Good to go :)

> If this works well I'll add this to the main repo as well

Can't wait to see how many alerts we get in the main repo

PR #1034, opened 2022-02-24
feat: run specific migration file with --name option

### LINKED ISSUE #860: Run Specific migration file
I can undo a specific migration file using `sequelize db:migrate:undo --name migration_file_name`.
I wonder why there is no option to migration just specific migration file?
Why below doesn't work?
`sequelize db:migrate --name migration_file_name`
It's just common sense everyone will go to this If they found out the method to undo a specific migration.

### LINKED ISSUE #1028: Sequelize migrations not working
## What you are doing?
Running cli with `--name` to undo specific migrations does not work anymore.

```js
npx sequelize-cli db:migrate:undo —-name 20220215204453-create-table

Sequelize CLI [Node: 14.18.1, CLI: 6.4.1, ORM: 6.14.1]

Options:
  --version          Show version number                                                                  [boolean]
  --help             Show help                                                                            [boolean]
  --env              The environment to run the command in                        [string] [default: "development"]
  --config           The path to the config file                                                           [string]
  --options-path     The path to a JSON file with additional options                                       [string]
  --migrations-path  The path to the migrations folder                             [string] [default: "migrations"]
  --seeders-path     The path to the seeders folder                                   [string] [default: "seeders"]
  --models-path      The path to the models folder                                     [string] [default: "models"]
  --url              The database connection string to use. Alternative to using --config files            [string]
  --debug            When available show various debug information                       [boolean] [default: false]
  --name             Name of the migration to undo                                                         [string]

Unknown arguments: —-name, 20220215204453-create-table
```

The migration undo runs smoothly

## What is actually happening?
I get an error that says `unknown arguments`

remarkablemark:
Closes #860, closes #1028

### Pull Request check-list

### Description of change

feat: run specific migration file with `--name` option

Example:

```sh
npx sequelize db:migrate --name 20111117063700-createPerson
```

### DIFF
--- a/src/commands/migrate.js
+++ b/src/commands/migrate.js
@@ -19,8 +19,10 @@ exports.builder = (yargs) =>
       type: 'string',
     })
     .option('name', {
-      describe: 'Migration name',
+      describe:
+        'Migration name. When specified, only this migration will be run. Mutually exclusive with --to and --from',
       type: 'string',
+      conflicts: ['to', 'from'],
     }).argv;
 
 exports.handler = async function (args) {
--- a/src/commands/migrate.js
+++ b/src/commands/migrate.js
@@ -17,6 +17,10 @@ exports.builder = (yargs) =>
     .option('from', {
       describe: 'Migration name to start migrations from (excluding)',
       type: 'string',
+    })
+    .option('name', {
+      describe: 'Migration name',
+      type: 'string',
     }).argv;
 
 exports.handler = async function (args) {
@@ -80,7 +84,13 @@ function migrate(args) {
           }
           return options;
         })
-        .then((options) => migrator.up(options));
+        .then((options) => {
+          if (args.name) {
+            return migrator.up(args.name);
+          } else {
+            return migrator.up(options);
+          }
+        });
     })
     .catch((e) => helpers.view.error(e));
 }
--- a/test/db/migrate.test.js
+++ b/test/db/migrate.test.js
@@ -595,6 +595,27 @@ describe(Support.getTestDialectTeaser('db:migrate'), () => {
         });
       });
     });
+
+    it('--name array', function (done) {
+      const migrationsPath = Support.resolveSupportPath('assets', 'migrations');
+      const migrations = fs.readdirSync(migrationsPath);
+      const createPersonMigration = migrations.find((migration) =>
+        migration.includes('createPerson')
+      );
+      const createPostMigration = migrations.find((migration) =>
+        migration.includes('createPost')
+      );
+
+      prepare(
+        `--name ${createPersonMigration} --name ${createPostMigration}`,
+        () => {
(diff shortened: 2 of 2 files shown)


src/commands/migrate.js
```diff
@@ -17,6 +17,10 @@ exports.builder = (yargs) =>
     .option('from', {
       describe: 'Migration name to start migrations from (excluding)',
       type: 'string',
+    })
+    .option('name', {
+      describe: 'Migration name',
```
ephys, line 22:
```suggestion
      describe: 'Migration name. When specified, only this migration will be run',
```

We should document that it's mutually exclusive with `to` & `from`.
We should also throw an error if these mutually exclusive options are provided together.

umzug can accept an array of migration name, we should make sure `name` can be specified more than once:

`sequelize-cli db:migrate --name migration1.js --name migration2.js`

ephys (changes requested):
Looking good and thank you for adding tests :)

src/commands/migrate.js
remarkablemark: c22a291

WikiRik:
Hi! Thanks for your work on this, could you also add a little bit about this to the [main sequelize documentation](https://github.com/sequelize/sequelize/blob/main/docs/manual/other-topics/migrations.md)? And maybe add a little bit to the description of the command here so it's clear to people that they can specify multiple migrations by repeating single migrations with the `--name` flag?

And as a bonus feature, can we make it so that `--name` accepts an array as well?

ephys (approved):
Looks good to me!

@WikiRik by array do you mean [this](https://github.com/sequelize/cli/pull/1034/files#diff-18525de583bcba92a422af366d5c414b0882d1d14e9092d07a398f89f6037498R610) or an alternative syntax?

WikiRik:
On mobile your link does not go anywhere unfortunately, but I was thinking about `--name createPersonMigration, createPostMigration` or something like that. It's a bit shorter than `--name createPersonMigration --name createPostMigration`.
But as I said, that's just a bonus. I think we can also merge without that and leave this for a future PR

ephys:
@WikiRik just checking: is this good for you? Can I merge?

WikiRik:
I don't see the comment anymore but we can merge this and add additional syntax options in a future PR

ephys:
I deleted my comment because I saw you already responded in the previous one 😅

adarsh-kumar1:
Not running as of now with single command using --
name

WikiRik:
True, we have merged this but not released a new version yet. We will do that in the coming days

SupunSPWithana:
> True, we have merged this but not released a new version yet. We will do that in the coming days

@WikiRik is this released now?

WikiRik:
> is this released now?

Yes, the latest v6 release includes this

PR #1046, opened 2022-03-11
Improve types.d.ts

arniu:
### Pull Request check-list

### DIFF
--- a/src/assets/migrations/create-table.js
+++ b/src/assets/migrations/create-table.js
@@ -1,6 +1,8 @@
 'use strict';
 
-/** @type {import('sequelize-cli').Migration} */
+/**
+ * @type {SequelizeCLI.Migration}
+ */
 module.exports = {
   async up (queryInterface, Sequelize) {
     await queryInterface.createTable('<%= tableName %>', {
--- a/src/assets/migrations/skeleton.js
+++ b/src/assets/migrations/skeleton.js
@@ -1,6 +1,8 @@
 'use strict';
 
-/** @type {import('sequelize-cli').Migration} */
+/**
+ * @type {SequelizeCLI.Migration}
+ */
 module.exports = {
   async up (queryInterface, Sequelize) {
     /**
--- a/src/assets/seeders/skeleton.js
+++ b/src/assets/seeders/skeleton.js
@@ -1,6 +1,8 @@
 'use strict';
 
-/** @type {import('sequelize-cli').Migration} */
+/**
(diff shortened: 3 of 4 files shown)


ephys:
Hi, thank you for the PR

Unfortunately I would need a good reason to merge this change. I don't want to declare a global namespace. I think importing the type is cleaner.

arniu:
These types are only referred in migration/seeder scripts. There is no need to import them explicitly.

ephys:
The fact that they're only referred in migrations is the reason why they're imported, instead of declared globally. It doesn't make sense to add something to the global object that is only used in a handful of files

It's also pretty standard https://github.com/ephys/new-sequelize-website/blob/ephys/docusaurus/docusaurus.config.js#L7

ephys:
projects have more than just migrations, so `sequelize-cli` is going to end up in the same package.json as the rest of the project's dependencies.

arniu:
The other source files  are different. 

There may be some dependencies and `import`s already. It's negligible to add a new one.

ephys:
I'm not sure I understand what you mean.

Nonetheless, the reason you gave me is not sufficient to add something to the global object. It's something that should be avoided as much as possible.

arniu:
@ephys 我改用中文，希望你能看懂。

## 会不会破坏已有代码？

`types.d.ts` 是最近加添加的，还没有发布，根本不会破坏已有代码。

## 为什么此处全局类型更合适？

migration 脚本跟其他代码不一样，它可以是某个工程的一部分，也可以是独立的脚本。

此时 `sequelize-cli` 就不再是一个 package，而是 migration 脚本的执行环境。作为上下文，执行环境导出最常见的全局类型，是合理的。

那么 `sequelize-cli` 会不会作为一个 package 使用呢？我看难，它也没导出其他东西。

ephys:
I'm hoping google translate gave me the right translation :)

It's true, this wasn't released yet so it's not a breaking change.

But whether migrations are standalone or not is not really relevant.
We don't have a good reason to declare a global type.
The only scenario where we would declare a global type is to type a global variable. But we don't have any.

It makes sense for libraries like `jest`, `mocha` etc, to do so because they expose their functions (`it`, `description`, `expect`) as globals.

Doing this in sequelize-cli would only impact the typing comment.
`/** @type {import('sequelize-cli').Migration} */`
would become
`/** @type {SequelizeCLI.Migration} */`

It's a few characters shorter but it's not worth declaring a global type.

PR #1141, opened 2022-08-19
Replace global process with local module

etroynov:
### Pull Request check-list

### Description of change

* Replace the global `process`  with a local module. According doc recomendation:
https://nodejs.org/api/process.html#process

* Bump node version 10 -> 16

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
   },
   "homepage": "https://github.com/sequelize/cli",
   "engines": {
-    "node": ">=16.0.0"
+    "node": ">=10.0.0"
   },
   "commitlint": {
     "extends": [
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "pg-hstore": "latest",
     "prettier": "2.7.1",
     "semver": "7.3.7",
-    "sequelize": "6.21.3",
+    "sequelize": "6.21.4",
     "sqlite3": "latest",
     "through2": "4.0.2"
   },
--- a/package.json
+++ b/package.json
@@ -20,9 +20,9 @@
     "yargs": "^16.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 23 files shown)


WikiRik:
We cannot bump the Node version without a major release. We have plans for v7 of the CLI (corresponding with v7 of sequelize itself) and it will most likely support Node 14 and above (more specifically 14.19.0 and higher iirc)

If this PR also makes sense with Node 10 we can review this

etroynov:
@WikiRik Thanks for the answer. I downgrade node version, can you review PR?

package.json
```diff
@@ -40,7 +40,7 @@
     "pg-hstore": "latest",
     "prettier": "2.7.1",
     "semver": "7.3.7",
-    "sequelize": "6.21.4",
+    "sequelize": "6.21.3",
```
WikiRik, line 43:
```suggestion
    "sequelize": "6.21.4",
```

src/assets/models/index.js
```diff
@@ -3,6 +3,7 @@
 const fs = require('fs');
 const path = require('path');
 const Sequelize = require('sequelize');
+const process = require('process');
```
WikiRik, line 6: Why is this not imported like the other files?

WikiRik (commented):
One small suggestion and a question.

Apart from those; can you also revert the changes in the `yarn.lock` file?

src/assets/models/index.js
etroynov: Because I thought it's not a good approach when you mix different import styles, but if you want a can change all `require -> import`.

WikiRik: Good point, let's leave this for a future refactor (which probably will happen when we work on the v7 version in TypeScript)

WikiRik (approved):

PR #931, opened 2020-10-14
Add db:info command which returns database configuration

reilg:
### Pull Request check-list

_Please make sure to review and check all of these items:_

- [x] Does `npm run test` pass with this change (including linting)?
- [ ] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?
- [ ] Have you added new tests to prevent regressions?
- [ ] Is a documentation update included (if this change modifies existing APIs, or introduces new ones)?

### Description of change

I wanted to add a way to quickly check if a command would be performed against the correct database connection.

Add the `db:info` command. 

Outputs:
```bash
Config {
  username: 'root',
  password: '***',
  database: 'test_database',
  host: 'localhost',
  port: '3306',
  dialect: 'mysql'
}
```

This isn't really very polished and no additional tests have been written to cover the command. I mainly wanted to open this PR to get comments and/or suggestions since I'm currently using this version for my current projects.

PR #1172, opened 2022-10-06
feat: added db:info command

reilg:
### Pull Request check-list

### Description of change

Adds the `db:info` command which displays the config information.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -19,3 +19,7 @@ insert_final_newline = true
 [*.md]
 trim_trailing_whitespace = false
 indent_size = 4
+
+[{Makefile,**.mk}]
+# Use tabs for indentation (Makefiles require tabs)
+indent_style = tab
--- a/config/config.json
+++ b/config/config.json
@@ -1,0 +1,36 @@
+{
+  "development": {
+    "username": "root",
+    "password": "secret_pass",
+    "database": "database_development",
+    "host": "127.0.0.1",
+    "dialect": "mysql",
+    "pool": {
+      "maxConnections": 1,
+      "some": {
+        "other": {
+          "nested": {
+            "object": "here"
+          }
+        }
+      }
+    }
+  },
+  "env": {
+    "use_env_variable": "postgresql://[user[:password]@][netlocation][:port][/dbname]"
+  },
+  "test": {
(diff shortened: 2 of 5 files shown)


WikiRik:
Thanks for the PR! I haven't had the time to take a look at this yet, but I haven't forgotten about it

WikiRik:
As further context, before I have to search for it again in Slack;

This PR was first opened as #931 

<details><summary>Suggestions from Slack, not very relevant</summary>
- A suggestion was made there to 'make the output a little prettier'
- There the suggestion was made to log it line by line instead of logging the object itself.

```
// Your example
Config {
  username: 'root',
  password: '***',
  database: 'test_database',
  host: 'localhost',
  port: '3306',
  dialect: 'mysql'
}

// Suggestion
Sequelize configuration:
username: root
password: ***
database: test_database
host: localhost
port: 3306
dialect: mysql
```

- An additional suggestion on Slack was to add a `--raw` flag so the output would still be the object
- A note added there was that the configuration can have nested objects and that those should work with this as well. A proposal for that was;

```
define.underscored: true
some.other.nested.value: value here
``` 

- The config used for the CLI tests would probably be used to test this new feature since it also has nested objects.

- Additionally a `--show-password` flag was proposed to show the password.
</details>

The tests that should be added here should use simplified versions of the [test config](https://github.com/sequelize/cli/blob/main/test/support/config/config.js) and check the output is as expected. Suggested tests are;
1. With a very basic config, verify that the output by using the command is correct
2. With a very basic config, verify that the output by using the `--raw` flag is correct
3. With a config with nesting, verify that the output by using the command is correct
4. With a config with nesting, verify that the output by using the `--raw` flag is correct
5. With a very basic config, verify that the output by using the `--show-password` flag is correct

These tests can be based on https://github.com/sequelize/cli/blob/main/test/version.test.js
One test that specifies a different config location that can be used is https://github.com/sequelize/cli/blob/main/test/options.test.js#L50-L66
Seeing that most/all tests should be run on both configs, something like the following could be useful to keep in mind; https://github.com/sequelize/cli/blob/main/test/db/seed/all.test.js#L7-L17

@reilg do you think you can work on this further with this information?

PR #1187, opened 2022-11-03
feat: add support for test files colocated with models

Lucianolo:
### Pull Request check-list

### Description of change

The models/index.js file is an auto-generated file used to automatically register all models from the models folder
It works under the assumption that the only files you have in your /models folder are models.
This works for most of the cases, but it causes an issue for projects using tests colocated with the tested file, e.g.
```
models/
    person.js
    person.test.js
    dog.js
    dog.test.js
    index.js
```
for such cases, if you want to add unit tests, you would have to either move to a symmetric /tests folder that replicates the structure of your project (for example as this project does), or follow this approach only for the models folder (quite inconsistent), or not write tests for models (quite bad)

This PR addresses the issue by adding a condition that will skip filenames containing `.test.js`.

### DIFF
--- a/src/assets/models/index.js
+++ b/src/assets/models/index.js
@@ -19,7 +19,12 @@ if (config.use_env_variable) {
 fs
   .readdirSync(__dirname)
   .filter(file => {
-    return (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js');
+    return (
+      file.indexOf('.') !== 0 &&
+      file !== basename &&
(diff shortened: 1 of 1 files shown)


Lucianolo:
i looked for a file with relevant test cases for this code but couldn't find it. If anyone knows where this logic is tested i'm happy to add additional test cases in there

WikiRik (approved):
I'm also not sure how to test this, but I think we can look into that later. Thanks for this PR!

PR #1241, opened 2023-01-25
fix: only show error detail when error.original is defined

Rochet2:
### Pull Request check-list

### Description of change

Using `"sequelize-cli": "6.5.2",`

When `up` handler of a migration throws a custom error.
```js
throw new Error(
  `foobar`,
);
```

The cli throws the following error
```bash
TypeError: Cannot read properties of undefined (reading 'detail')
    at Object.error (/myproject/node_modules/sequelize-cli/lib/helpers/view-helper.js:31:24)
    at /myproject/node_modules/sequelize-cli/lib/commands/migrate.js:68:39
    at async exports.handler (/myproject/node_modules/sequelize-cli/lib/commands/migrate.js:27:7)
```

The error originates from this line:
```js
    if (error.original.detail) { // this causes the error because original is null/undefined
      console.error(`${_cliColor.default.red('ERROR DETAIL:')} ${error.original.detail}`);
    }
```

In this PR we change the IF condition to check if the original field is set so the error will not occur.

### DIFF
--- a/src/helpers/view-helper.js
+++ b/src/helpers/view-helper.js
@@ -37,7 +37,7 @@ module.exports = {
 
     this.log();
     console.error(`${clc.red('ERROR:')} ${message}`);
-    if (error.original.detail) {
+    if (error.original?.detail) {
       console.error(`${clc.red('ERROR DETAIL:')} ${error.original.detail}`);
     }
 
--- a/src/helpers/view-helper.js
+++ b/src/helpers/view-helper.js
@@ -37,7 +37,7 @@ module.exports = {
 
     this.log();
     console.error(`${clc.red('ERROR:')} ${message}`);
-    if (error.original?.detail) {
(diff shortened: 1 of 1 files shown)


WikiRik (approved):
Thanks for this PR!

src/helpers/view-helper.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
 
     this.log();
     console.error(`${clc.red('ERROR:')} ${message}`);
-    if (error.original.detail) {
+    if (error.original?.detail) {
```
WikiRik, line 40:
```suggestion
    if (error.original && error.original.detail) {
```

WikiRik (changes requested):
Sorry, approved too quickly. We still support Node 10 which does not have the optional chaining

WikiRik (approved):
Thanks for the quick update!

Issue #1436, opened 2023-12-25
When on typescript project, sequelize-cli still generates javascript files.


malik-samad:
## What I am doing?
I'm using Sequelize-cli in a typescript project and it is not generating .ts files and I have to then update those files extensions and make manual changes to the file for making it compatible in my project.

## What I expect to happen?
I want to have the CLI work the same for my typescript file as it does for a javascript project. I expect the cli to generate proper migration, see and model files with .ts extension.

## What is actually happening?
When I try to run `sequelize init` or other migration cli, it will generate javascript files instead of typescript as expected.

Dialect: mysql
Database version: latest
Sequelize CLI version: latest
Sequelize version: latest

WikiRik:
Thanks for opening this! You can fork this repository, push the changes to there and then open a pull request here

berubenic:
+1

malik-samad:
Hi WikiRik, 
I have opened a PR for this issue and it needs your review. please have a look and help me get in making it merged into package.

The PR has 1 regex change which is reported by CodeQL and it needs approving from reviewer as I think its a false positive.

thanks.

IgorArnaut:
I wish there was an option to generate TypeScript version of the project.

@malik-samad
A year later and your pull request is still not merged.

omertairi04:
Why exactly isn't this merged ?

malik-samad:
> Why exactly isn't this merged ?

Hi, thanks for checking.
The reason is because the PR needs approval from the reviewer. I can't merge it on my own.

omertairi04:
> > Why exactly isn't this merged ?
> 
> Hi, thanks for checking. The reason is because the PR needs approval from the reviewer. I can't merge it on my own.

Since i am a beginner when it comes to things as such, is there any work around to implement your solution ?

shlomisas:
@WikiRik any update on ETA to approve the PR? 🙏

IgorArnaut:
> [@WikiRik](https://github.com/WikiRik) any update on ETA to approve the PR? 🙏
There's no ETA. The main devs just don't care.

malik-samad:
> > > Why exactly isn't this merged ?
> > 
> > 
> > Hi, thanks for checking. The reason is because the PR needs approval from the reviewer. I can't merge it on my own.
> 
> Since i am a beginner when it comes to things as such, is there any work around to implement your solution ?

Yes, You can actually check my other repo for this implementation.

[...]

PR #1439, opened 2023-12-25
feat: configuration for seamless cli use in a typescript project

malik-samad:
- In this commit, I have modified the templates and made other related changes to add support for typescript projects.

- The new changes I made, will automatically detect if the project that uses sequelize-cli is on typescript or not and will switch logic based on it.

### Pull Request check-list

### Description of change

Link to the issue: https://github.com/sequelize/cli/issues/1436

The sequelize-cli didn't have the support for typescript projects and in this PR I implemented certain changes that will make it capable of handling the CLI for both typescript and javascript files, additionally, I have improved the templates to be able to generate both `.ts` and `.js` files for migrations, models, and seeds.


src/helpers/model-helper.js
```diff
@@ -24,10 +56,13 @@
       fieldName: split[0],
       dataType: split[1],
       dataFunction: null,
+      tsType: getTsTypeForDbColumnType(split[1]),
       dataValues: null,
     };
   } else if (split.length === 3) {
-    const validValues = /^\{(,? ?[A-z0-9 ]+)+\}$/;
+    const validValues =
+      /^\{((('[A-z0-9 ]+')|("[A-z0-9 ]+")|([A-z0-9 ]+)))(, ?(('[A-z0-9 ]+')|("[A-z0-9 ]+")|([A-z0-9 ]+)))*\}$/;
```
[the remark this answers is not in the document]
malik-samad: it's a false positive, @WikiRik can you please dismiss this alert?

src/helpers/import-helper.js
```diff
@@ -12,6 +12,19 @@ async function supportsDynamicImport() {
...
+    // Try to require the package
+    require.resolve(packageName);
+    return true;
+  } catch (error) {
+    // If require.resolve throws an error, the package is not installed
+    return false;
+  }
+}
+
+const isTypescriptProject = isPackageInstalled('typescript');
```
shlomisas, line 26: what about having a way to decide it from the CLI? e.g. `npx sequelize-cli migration:generate --name XXX --typescript`

WikiRik: This is also how we are going to do it in the new CLI; https://github.com/sequelize/sequelize/blob/main/packages/cli/src/commands/migration/generate.ts

PR #1551, opened 2025-11-05
fix: move to trusted publishing

WikiRik:
### Pull Request check-list

- [X] Does `npm run test` pass with this change (including linting)?
- [X] Does the description below contain a link to an existing issue (Closes #[issue]) or a description of the issue you are solving?

### Description of change

This PR moves our release flow to use [trusted publishing](https://docs.npmjs.com/trusted-publishers/). Also moves the CI workflow to npm and updates CI in general. Moves the release workflow to a separate one that is called when we want to do a release. This manual action is in line with how we do `@sequelize/*` releases, I have removed the tests from it since tests need to pass before we merge.

About the changes to notifications; twitter was not working so I've removed that. The special OpenCollective action was nice, but did a lot that I feel like we can just include it directly (especially with v7 being a monorepo so we only define it once there)

Marked as a fix so it will publish a new release


sdepold (approved):

.github/workflows/ci.yml
```diff
@@ -1,96 +1,67 @@
 name: CI
-on:
-  push:
-    branches:
-      - main
-      - v6
-      - v7
-  pull_request:
+on: pull_request
```
sdepold, line 2: why does this migrate from on push to on pull_request?

.github/workflows/ci.yml
```diff
@@ -1,96 +1,67 @@
...
           node-version: ${{ matrix.node-version }}
-      - run: yarn install --frozen-lockfile --ignore-engines
-      - run: yarn add sequelize@${{ matrix.sequelize-version }} --ignore-engines
+      - run: if [ "${{ matrix.node-version }}" = "10" ]; then npm install -g npm@7; fi
+      - run: npm ci
+      - run: npm install sequelize@${{ matrix.sequelize-version }}
       - run: docker compose up -d ${DIALECT}
       - run: docker run --link ${DIALECT}:db --net cli_default jwilder/dockerize -wait tcp://${DIALECT}:${SEQ_PORT::-1} -timeout 2m
-      - run: yarn test
+      - run: npm test
```
sdepold, line 26: while i dont mind, why are we moving to npm=

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Post release notes to OpenCollective
+        if: steps.release.outputs.new_release_published == 'true'
+        env:
+          OPENCOLLECTIVE_TOKEN: ${{ secrets.OPEN_COLLECTIVE_KEY }}
+          OPENCOLLECTIVE_SLUG: sequelize
+          RELEASE_VERSION: ${{ steps.release.outputs.new_release_version }}
+          RELEASE_NOTES: ${{ steps.release.outputs.new_release_notes }}
+          PACKAGE_NAME: sequelize-cli
+        run: |
+          curl -X POST "https://api.opencollective.com/graphql/v2" \
```
sdepold, line 43: does this mean we dont need the dispatch job anymore that i created a few years ago?

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+            -H "Content-Type: application/json" \
+            # TODO: use OAuth instead of Personal-Token so we can create the updates from the organization instead of a user
```
sdepold, line 45: what about this?

package.json
```diff
@@ -120,21 +120,8 @@
     "*.js": "eslint"
   },
   "release": {
-    "plugins": [
```
sdepold, line 123: you are really removing a lot :D why is that?

.github/workflows/ci.yml
WikiRik: We don't run it 'on push' at the regular repo, so this makes it more consistent. Reasoning there was because we do everything with PRs that are up to date, we don't need to run it again after merging

.github/workflows/ci.yml
WikiRik: Because yarn doesn't offer the trusted publishing (yet). So we're locking in to the npm registry and its package manager

.github/workflows/release.yml
WikiRik: Yup!

.github/workflows/release.yml
WikiRik: Now when we post something to OpenCollective it's done under your name. With OAuth we should be able to do it under the name of the organization. This is a TODO for a later PR

package.json
WikiRik: It's the default config already, so this cleans it up

